Issue #165, opened 2026-09-13
Dial plan has no outbound-trunk routing action, and no E.164 normalization anywhere


GlomarGadaffi:
## Gap

Confirmed by reading the dial-plan engine (\`DialPlan.hpp\`, \`PbxFeatureConfig.cpp\`'s
\`routeDialPlan\`, \`CallForker.cpp\`) and the anchor-call wiring
(\`RequestsHandler.cpp\`'s \`kAnchorCallExt\`):

- The dial plan's action set today is internal-routing only: match a pattern
  (literal / \`X\` digit-wildcard / trailing \`*\` prefix-wildcard), then \`group\`
  (ring group), \`page\` (paging zone), or \`park\` (park orbit). There is **no
  \"route to trunk\" action.**
- The only way to reach the PSTN via the ported \`TelephonyAnchorClient\` today is
  dialing the literal fixed virtual extension \`555\` (\`kAnchorCallExt\`,
  RequestsHandler.cpp:72) — an internal extension has to know and dial \"555\", not
  the actual number they want to reach. There's no path where dialing an ordinary
  external-looking number (\`+15551234567\`, or \`915551234567\`, or whatever the site's
  outbound convention is) gets automatically routed to the trunk.
- There is **no E.164 normalization utility anywhere in the codebase**
  (\`grep -rn \"e164\|E164\" src/\` — zero hits). Every AOR/number comparison in the
  engine (registrar lookups, dial-plan pattern matching, CDR caller/callee fields)
  operates on whatever literal string arrived, with no normalization pass to
  reconcile \`+15551234567\` / \`15551234567\` / \`5551234567\` as the same destination.

## Why this matters

This is the concrete blocker behind two other things:
- **911 dialing** (tracked separately) has nowhere to route to without a trunk
  action existing at all — \"dial 911, get routed to whatever PSTN trunk is active\"
  requires the dial-plan engine to have an outbound-trunk concept in the first
  place.
- **ITSP peering** (tracked separately) needs a place to plug into once it exists —
  a generic SIP trunk is exactly the kind of destination the dial plan's routing
  actions should be able to target, the same way it already targets ring groups,
  page zones, and park orbits.

## Shape of the work (tracking placeholder, not a spec)

1. A new dial-plan action, e.g. \`trunk\` — routes a matched pattern to whichever
   provider the active \`TelephonyApiConfig\` slot (or, once #164 lands, the active
[...]

Issue #94, opened 2026-07-18
docs/API.md endpoint catalog is stale — missing CDR/DND/forward/group/OTA/factory-reset specs


GlomarGadaffi:
docs/API.md's detailed endpoint catalog (§4–§5) still specifies only the original five endpoints (`/api/status`, `/api/kill`, `/api/wifi/scan`, `/api/wifi/connect`, `/api/wifi/mode_ap`).

Endpoints shipped since that are **not** fully specified: `/api/cdr`, `/api/dnd`, `/api/forward`, `/api/group`, `/api/factory-reset`, `/api/configuring`, `/api/ota/status`, `/api/ota/reboot`, the OTA upload path, and the admin session endpoints.

A summary of the admin layer + dark-by-default reachability was added as API.md §0 (2026-07-17) with a partial-catalog warning banner, so the doc is no longer *misleading* — but the full per-endpoint specs (request/response schemas, status codes, auth requirements) still need writing. Source of truth: `src/Helpers/HttpServer.cpp` `handleClient()` dispatch.

Issue #178, opened 2026-09-13
Provisioning: DHCP Option 66 zero-touch discovery


GlomarGadaffi:
> Ported from drawbridge #136. Verified gap here: stock `esp_netif_dhcps_start` only (`esp_main.cpp:134`, `esp_main_eth.cpp:443`); no Option 66 injection.

Phones discover their provisioning URL from DHCP Option 66 (and 160/242 on some vendors). For the SoftAP / built-in DHCP server case, inject Option 66 = `http://<ip>/config/` — the IDF `dhcpserver` needs a patch or the newer custom-option hook (check the IDF 6 API before forking). For wired LANs, document configuring the site DHCP server instead. Pairs with #158/#138.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01JJSmyBFe3aesr4id9pEMm7

GlomarGadaffi:
Surfacing this as a comment (found during tonight's tracker-hygiene sweep) since the issue body already contains the actual investigation conclusion but it's never been called out as one: the fix approach here is settled, not still an open question. Concretely — inject DHCP Option 66 = `http://<ip>/config/` from the SoftAP/built-in DHCP server path, which needs either a patch to IDF's `dhcpserver` or its newer custom-option hook (check the IDF 6 API surface at implementation time, since this was written against an earlier IDF version's docs). Wired LANs get documentation pointing at the site DHCP server instead of firmware-side injection — no code path needed there.

Pairs with #158/#234 (the multi-vendor renderer wiring) — Option 66 is only useful once a phone is actually being told which config format to fetch.

No scope change, just making the conclusion visible in the timeline for whoever picks this up.

Issue #148, opened 2026-09-07
Register-beep INVITE retransmits ~55x in 8s and continues after its own CANCEL, exhausting the message pool


GlomarGadaffi:
Found on real hardware (LilyGO T-ETH-Elite S3, W5500, `SIP_TRANSPORT=eth`) while verifying #103.

## Observed

A single REGISTER from one client triggers the register-beep INVITE, which then retransmits far outside RFC 3261's timer A/B schedule. Every packet the board sent to one client in an 8-second window, captured raw:

```
[1]  INVITE sip:1001@192.168.12.110:5075   Call-ID: 6RsZEFqQrf7SKtev@192.168.12.244  CSeq: 1 INVITE
[2]  SIP/2.0 200 OK                        Call-ID: regdiag-opbc7mhs                 CSeq: 1 REGISTER   <- our REGISTER, fine
[3]  INVITE ... same Call-ID ...
[4]  OPTIONS sip:1001@...                  (keepalive, fine)
[5..25]  INVITE ... same Call-ID ...       (21 more)
[26] CANCEL sip:1001@192.168.12.110:5075   Call-ID: 6RsZEFqQrf7SKtev  CSeq: 1 CANCEL
[27..59] INVITE ... SAME Call-ID ...       (33 MORE, after the CANCEL)
```

**~55 INVITE retransmissions in 8 seconds**, on one Call-ID, and **retransmission continues after the server sent its own CANCEL** for that transaction.

RFC 3261 §17.1.1.2: a UAC INVITE transaction retransmits on timer A with exponential backoff (T1, 2×T1, 4×T1…) and gives up at timer B = 64×T1 — **7 retransmits over 32 seconds**, not 55 in 8. And once CANCEL is sent, the INVITE should stop being retransmitted.

## Consequence

This is the direct cause of the `[tx] pool exhausted — INVITE sent without retransmit tracking` messages, which appear in bursts on the console:

```
52 [tx] pool exhausted — INVITE sent without retransmit tracking
```

Under a 45-client registration storm the pool stays starved for the whole run. A starved pool is not cosmetic — call paths explicitly drop work when `getMessageFromPool()` returns null (`if (!response) return;   // pool exhausted: drop, peer retransmits (#101A)`), so unrelated in-flight signalling gets silently dropped while the beep floods.

## Repro

1. Flash any current `esp32s3-eth` build, provision it so the SIP stack comes up.
[...]

PR #187, opened 2026-09-14
Dial plan, groups, call log and trace as toolbar modals + three silent dial-plan/trunk fixes

GlomarGadaffi:
## Why

A live board with a real Yealink T29 registered could not place an outbound call over the 3CX trunk. Tracing that turned up one missing UI and three silent defects — silent in the strict sense: nothing logs, nothing errors, and the dashboard shows a healthy box the whole time.

The starting symptom was `"dialplan":[]` on a board with ~2 hours of continuous uptime. That is not a misconfiguration so much as a dead end: the dial plan is the **only** route to an outside trunk (there is no hardcoded `9` prefix and no unregistered-destination fallback — drawbridge has both, this engine has neither), and there was **no dashboard UI for dial rules at all**. Every rule had to be hand-`POST`ed, and the only in-page hint was a note reading `POST /api/dialplan (see docs/API.md)`.

### 1. Toolbar: Dial Plan · Groups · Call Log · Trace (`c0b417e`)

Ring Groups & Forwarding, Call Log and SIP Trace were inline rack modules; the dial plan had no surface at all. All four are now top-bar buttons beside the existing ones:

```
Refresh · Dial Plan · Groups · Call Log · Trace · WiFi · Admin · Interconnect · Help
```

`F2`/`F3`/`F4`/`F8` join `F1`/`F5`/`F9`, and `Escape` closes all nine. `<main>` is now just the patch bay.

The three existing modules moved into `.modal.wide` overlays with **every control id preserved**, so `renderGroups()`, `renderCdr()` and the trace terminal keep working untouched.

**Dial Plan is new.** It lists the rules from `/api/status`, renders each as `9XXXXXXXXXX → trunk → strip 1, prepend 1` with a Delete button, and hides the strip field for non-trunk actions.

Also re-split the page **7 → 8 parts**. Adding the modal pushed `PD_HTML_1` to 17,233 bytes, past the ~16 KB single-literal cap MSVC enforces — that would have broken the host build on `cl`. Parts are now ~12 KB each and reassembly was verified byte-identical.

### 2. Three fixes (`64cfb6d`)

**`DialPlan::upsert()` dropped `stripDigits` on an edit.** It copied only `action` and `target` when the pattern already existed, so `stripDigits` kept the *first* insert's value forever — and `persistDialPlan()` then wrote that stale value to NVS, where it survived a reboot. Since re-`POST`ing a pattern is the only way to edit a rule, this hit the normal editing path: a rule saved as strip 1 / prepend 1 and later edited to fix a typo would dial `193057673260` instead of `13057673260`.

**"Strip N, prepend nothing" could not be expressed.** The delete signal was an empty `target`, which made the request that *creates* a strip-only trunk rule byte-identical to the request that *deletes* it. That shape is not exotic — it is exactly what drawbridge sends on this same trunk (a bare `substr(1)`, no prepend), so the one digit format known to work here was the one format pocket-dial could not store. Delete is now **empty `action` AND empty `target`**; naming an action always means upsert. Every existing delete call site already passes `("pattern", "", "")`, so nothing changes for them. A non-trunk action with an empty target is now refused outright instead of silently deleting the rule being edited.

**An enabled trunk slot saved with a blank route DN.** `setSlot()` checked field lengths and the `https://` prefix and nothing else. A blank `routeDn` is the one misconfiguration nothing downstream can report: `isConnected()` is WebSocket state only and the control socket carries no DN, so the board boots "connected", every `makeCall` POSTs to `.../callcontrol//makecall`, the provider is genuinely reached and answers non-2xx. In the field that reads as *"the API was hit but no call happened."* An enabled non-Loopback slot now requires a route DN, a client ID and a secret — the validation drawbridge's trunk editor has always had (`"Client ID, secret and source DN are all required"`) and that pocket-dial lost when it replaced `TrunkConfig` with the slot model. The secret check honours `keepSecret`, so editing a saved slot still works, and disabled slots stay free to be incomplete drafts.

## Verification

- **408/408 host tests pass** (405 before — three new).
- Each fix has a regression test, and **each test was verified to FAIL against the unfixed code** before being kept. The telling one: on the old code a blank-`routeDn` slot came back with `view().enabled == true`, so it really did save and activate.
- Host build clean under gcc; only the two pre-existing `-Wc++20-extensions` warnings.
- Dashboard rendered live in Chrome: all nine buttons, all four modals opening, `renderDialplan()` exercised with real-shaped data, **zero console errors**. Tag balance, overlay wiring, `$()` id references and the `PD_CSRF` placeholder all checked by script.

**Hardware: untouched.** Nothing here has been flashed to a board.

## Notes

- `docs/API.md`'s one-line `/api/dialplan` entry now states the parameters and the delete signal. Its `#post-apidialplan` anchor is still dangling — that section never existed, which is pre-existing (#94), not introduced here.
- The trunk investigation that prompted this is not finished: the outbound call still needs a live retest once a rule exists, and 3CX's own outbound rule for this DID has an empty **Calls from Departments** field that is worth confirming.

Refs #165, #69

https://claude.ai/code/session_01SGVsyCvpmPTAEXSGjMHe52


GlomarGadaffi:
## Hardware verification — outbound PSTN works end to end

Bench: LilyGO T-ETH-Elite S3 (`SIP_TRANSPORT=eth`), real Yealink T29 as ext 1001, 3CX tenant, route DN `rcv2`. Dial rule `9XXXXXXXXXX -> trunk 1, strip 1` entered through the new Dial Plan modal.

**Call 1 — rang through to carrier voicemail:**
```
t=110505  Successfully initiated call to 1305******0 (own leg 25)
t=131702  Upset 25 -> control leg 25 status 'Connected'
          POST (device->Telephony) audio stream OPEN
          [Telephony] Event: Answered, participantId=25
```
Voicemail answered at **21.2 s**. The previous budget was `pbx::kNoAnswerTimeout` (20 s) measured from INVITE-receipt, i.e. ~19 s of real ring after the makeCall round trip — it was reaping the call **~2 s before** voicemail picked up. That is the defect, with a number.

**Call 2 — answered by a person, two-way audio confirmed:** `Dialing -> Dialing -> Connected` at 6.7 s, POST stream open, `Answered`.

**Resolves the open question in the commit message:** 3CX *does* emit a `Connected` upset for a makecall-originated leg. The synthetic-upset poll considered as a possible fallback is **not needed**.

## Register-beep storm (#148) — fixed, verified

Same board, same phone, same fresh-boot new binding that produced the failure:
```
New Client: 1001
Register beep: INVITE -> 1001
Register beep: no answer from 1001, cancelled
```
Before: **23** `[tx] Timer B expired` for one Call-ID plus `[tx] pool exhausted`. After: **zero and zero**; serial log 595 bytes vs 5759. The single `Timer B expired` seen later in the log is RFC 3261 §17.1.1.2-correct — one unanswered INVITE, one transaction, one expiry.

**Residual, benign:** the phone answers our CANCEL with `481 Call Leg/Transaction Does Not Exist`, which has no dispatch-table entry and logs as `[SIP] ERROR`. That is the gap drawbridge closed in its #178 with `onFinalFailure()` — deliberately not ported here, and a good follow-up.

Issue #188, opened 2026-09-14
Factory reset erases reg_mode from the wrong NVS namespace — the secure-mode lockout rescue #151 describes does not work


GlomarGadaffi:
## Summary

`DeviceConfig::clearAll()` erases `reg_mode` from the **wrong NVS namespace**, so a factory reset does not clear the registrar admission mode. This is the same namespace-mismatch class as #151 — and it is the half of #151 that never got fixed.

## The mismatch

`src/Helpers/DeviceConfig.cpp:628` opens the `storage` namespace and erases the key there:

```cpp
constexpr const char* kNvsNamespace = "storage";   // :49
constexpr const char* kKeyRegMode   = "reg_mode";  // :53
...
if (nvs_open(kNvsNamespace, NVS_READWRITE, &h) == ESP_OK)   // :628  <-- "storage"
{
    ...
    nvs_erase_key(h, kKeyRegMode);                          // :642
```

But the registrar reads and writes it in `pbxcfg`:

```cpp
// src/SIP/Registrar.cpp:35
if (nvs_open(pbxpersist::kNvsNamespace, NVS_READWRITE, &h) != ESP_OK)   // "pbxcfg"
...
esp_err_t err = nvs_get_u8(h, "reg_mode", &v);                          // :40
```

So the erase targets a key that, on a v1.4.1+ board, does not exist in that namespace. The registrar's value in `pbxcfg` survives the reset untouched.

## Why this matters more than a stray key

The code's own comment at `DeviceConfig.cpp:635-641` states exactly what this erase is for:

> The registrar admission mode goes too. This key belongs to Registrar, but DeviceConfig is what writes it from the flash seed, and leaving it behind made factory reset unable to rescue the one state that most needs rescuing: a device switched to `secure` before any extension was secured rejects every REGISTER, locking every phone (and thus the operator) out. Without this, the only way back was USB.

That rescue does not work. A board flipped to `secure` before any extension was adopted still rejects every REGISTER **after** a factory reset, and the only way back really is USB — the precise outcome the code was written to prevent.

## Why #151's guard did not catch it

#151 fixed the **write** path and added a named constant plus a compile-time guard:

- `src/Helpers/DeviceConfig.hpp:163` — `constexpr const char* kRegistrarNvsNamespace = "pbxcfg";`
- `src/Helpers/DeviceConfig.cpp:567` — `applyFlashSeed()` correctly uses `kRegistrarNvsNamespace`
[...]

Issue #151, opened 2026-09-07
cfgseed regMode is written to NVS namespace "storage" but Registrar::loadMode() reads "pbxcfg" — flash-time registrar mode has never worked


GlomarGadaffi:
## Summary

`DeviceConfig::applyFlashSeed()` writes the seed's `regMode` to NVS key `reg_mode` in namespace **`"storage"`**. `Registrar::loadMode()` reads `reg_mode` from **`pbxpersist::kNvsNamespace`, which is `"pbxcfg"`** (`src/SIP/PbxPersist.hpp:16`).

Different namespaces, so the value never arrives. The flash-time registrar mode has **never worked on any release**.

The operator-facing consequence is the security-relevant part. `docs/LEARN_MODE.md`, `docs/INCIDENT_PLAYBOOK.md` §"registrar modes" and `DeviceConfig.hpp`'s own comment all state that the `cfgseed` `regMode` field is *the only* way to put a **headless** `esp32s3-eth` / `esp32s3-wifi` board into `learn` or `secure` before first boot. It is not — it does nothing. And once such a board is provisioned the HTTP admin plane is dark, so the only remaining route is the `*4887` DTMF star-code, which is unreliable (it fired once for me and would not reproduce, on a byte-identical build). A headless board is effectively stuck in `open`.

This was masked until now by a second bug: the manifest generator's byte-swapped partition magic (fixed in `e559368`, v1.4.0) meant `cfgseed_offset` was never emitted, so the browser flasher never wrote a seed at all and `regMode` never got the chance to fail. Fixing the outer bug exposed the inner one.

## Where it is

```cpp
// src/Helpers/DeviceConfig.cpp:46
constexpr const char* kNvsNamespace = "storage";
...
// applyFlashSeed(), regMode branch -- writes into "storage"
if (regMode <= 2 && nvs_set_u8(h, "reg_mode", regMode) == ESP_OK)
```

```cpp
// src/SIP/PbxPersist.hpp:16
constexpr auto kNvsNamespace = "pbxcfg";

// src/SIP/Registrar.cpp:30
void Registrar::loadMode()
{
    nvs_open(pbxpersist::kNvsNamespace, NVS_READWRITE, &h);   // "pbxcfg"
    esp_err_t err = nvs_get_u8(h, "reg_mode", &v);
```

[...]

PR #210, opened 2026-09-14
feat(nvs): stamp a schema version at boot, so a future release can migrate

### LINKED ISSUE #181: NVS schema versioning and migration framework
> Ported from drawbridge #133. Verified gap here: no `schema_version`/`cfgVersion`/`migrat` in `src/`.

The `pbxcfg` key space grows ad hoc (dial plan, ring groups, park, provisioning, anchor creds, soon SMTP). A rename silently reads stale data from the old key and empty from the new one — undiagnosable in the field.

1. `schema_ver` (uint16) written at first boot / first provisioning.
2. `migrations[] = { from, to, fn }` run in order on boot when `schema_ver < CURRENT`.
3. Each migration commits atomically; on failure log and boot with defaults for the affected keys rather than crashing.
4. Config export (see the export/import issue) embeds `schema_ver`; import refuses newer, migrates older.

Do this **before** the first export format ships.

https://claude.ai/code/session_01JJSmyBFe3aesr4id9pEMm7

GlomarGadaffi:
Closes #181

## The problem

This device keeps its whole identity in NVS — the admin credential, the SIP extensions and their secrets, the dial plan, ring groups, the DID map, the trunk API slots with their secrets, WiFi config, the syslog target — and firmware is updated over the air. **Nothing recorded which layout those keys were written in.** The first release to change what a key *means* therefore had exactly two options, and both are bad: misread the old value in silence, or tell the operator to factory-reset a live phone system and re-provision it by hand.

## What lands

`DeviceConfig::ensureSchemaVersion()` runs at boot in all four `main/esp_main*` entry points, immediately after `nvs_flash_init()`. It probes NVS, decides what layout the device is holding, runs whatever migrations that implies, stamps the result, and logs what it did.

It is shaped as a **sibling of `applyFlashSeed()`** rather than as a new idea: both are "inspect persistent state once at boot, act exactly once, record that you acted so the next boot does not repeat it". `cfgseed_gen` is that counter for the flash seed; `schema_ver` is it for the key layout.

**This release stamps the current layout as version 1 and ships an empty migration table.** There is no earlier layout to convert from, and inventing rows for changes that have not happened would be untested flash-mutating code in production. The dispatch is fully exercised by host tests against synthetic tables, so adding the first real row is a two-line change to one function.

## Ordering: before `applyFlashSeed()`, not after

The call is inserted **above** the seed block in every entry point, and that is load-bearing. The seed writer opens `storage` and `pbxcfg` `NVS_READWRITE`, which **creates** them. Run it first and the "does this device already hold config?" probe finds namespaces the seed itself just made, concludes every board is brand new, and skips the migrations a real device needs. For the same reason every probe is `NVS_READONLY` — `ESP_ERR_NVS_NOT_FOUND` from a READONLY open is the only way to ask "does this namespace exist" without answering the question by creating it.

## Deviation from the issue: `storage`, not `pbxcfg` — please read this first

The issue frames this as "the `pbxcfg` key space grows ad hoc", so `pbxcfg` is the obvious home. **I put the stamp in `storage` instead**, and a reviewer will read that as a conflict unless I explain it:

1. The version describes the **whole device**, not one namespace. A firmware image upgrades every namespace it owns at the same instant — `storage`, `pbxcfg`, `sipauth`, `didmap`, `tapicfg` — so six per-namespace stamps would be six chances to disagree about a state that can only ever have one value.
2. More concretely: reaching into `pbxcfg` from `DeviceConfig.cpp` is **the exact move that produced #151 and then #188**. The scar tissue is thirty lines above the new code in this very file. `storage` is the namespace `DeviceConfig` already owns, it exists on every transport including the pure-Ethernet ones, and putting the stamp there needs no cross-namespace exception at all.

The issue does not actually name a namespace, so this is a judgement call rather than a contradiction — but it is the one design point I would most like challenged.

## The decision table

| Probe | Outcome | Action |
| :--- | :--- | :--- |
| no stamp, no config namespace exists | `FreshInstall` | stamp current |
| **no stamp, some config namespace exists** | **`AdoptedLegacy`** | **read as v1, migrate forward, stamp. NEVER wipe** |
| stamp == current | `UpToDate` | nothing written at all |
| stamp < current | `Migrated` | walk the table, re-stamping after **each** step |
| stamp > current | `Downgrade` | nothing written, loud `ESP_LOGE`, still boots |
| store not inspectable | `StoreUnavailable` | nothing written |

The row in bold is every board already in the field on the release that introduces versioning. Its data **is** v1 — that is a fact about what every shipped release wrote, not a guess. Wiping there would factory-reset the deployed fleet on the update that added versioning, which is worse than the problem.

"Cannot tell" deliberately never collapses into "nothing there": stamping a provisioned device as a fresh install is precisely the silent skip this mechanism exists to prevent, and a transient `nvs_open` failure would otherwise cause it.

A stored version of `0` is treated as **no stamp** rather than as version zero. No release has written 0, so reading one means a foreign or corrupt write; the legacy path re-stamps it with a real number instead of trying to migrate "from version 0", for which no table row can ever exist.

## Why a downgrade logs loudly and still boots

Refusing to start was considered and rejected. Old firmware meeting a newer stamp is **not** a hypothetical that needs an operator to cause: OTA rollback is armed (`CONFIG_BOOTLOADER_APP_ROLLBACK_ENABLE`, `main/esp_main.cpp:235`) and the new image is only marked valid after several seconds of healthy operation, so "new firmware migrated to v2, stamped it, then crashed" hands an old image a v2 store automatically. Refusing to boot would turn the anti-crash safety net into the brick. A PBX that will not ring is worse than a PBX running on defaults.

Reading the keys anyway and saying nothing was also rejected — that is the silent misparse this whole change exists to stop.

So: detect, log at `ESP_LOGE` with an unmissable two-line banner, write **nothing** (stamping backwards would let the newer firmware return and mistake its own v2 data for v1), migrate nothing, and boot. The newer data is left intact, so re-flashing the newer image restores the device exactly.

**Known gap, stated rather than hidden:** this does not quarantine the store against the old firmware's own *writes*, so v1-shaped values can be written into a v2-stamped store. Gating every loader and writer on the outcome is a much larger change and there is no v2 to test it against. `lastSchemaOutcome()` / `observedSchemaVersion()` are exposed so that work builds on this rather than redesigning it.

## Per-step commit

`runSchemaMigrations()` takes a stamp callback and invokes it after **every** successful step rather than writing the final version once at the end — a power cut between two conversions must leave the device at the last layout that actually landed. A migration that converted the data but could not record that it did aborts the chain. A missing row is a hard failure, never a skip. A row that does not move forward is refused *without being executed*, and the walk is bounded by the table size, so a hand-written cycle is a failed migration rather than a board that never finishes booting.

## `clearAll()` does not drop the stamp

Factory reset clears three keys out of `storage` and leaves `pbxcfg`/`sipauth`/`didmap`/`tapicfg` alone, so the device still holds config in the current layout afterwards. Erasing the stamp would make the next boot see "data, no stamp", conclude v1, and on a future v3 firmware re-run the 1→2→3 migrations over data that is already v3. The stamp describes the layout, not the contents.

## Testing

`planSchema()` and `runSchemaMigrations()` are pure, platform-neutral, and take plain function pointers (no `std::function`, no captures, no allocation), so they compile and run on the host where there is no NVS. Only the `nvs_open`/`nvs_get_u16`/`nvs_set_u16` glue is inside the ESP guard. **That split is the #151 lesson applied ahead of time:** logic buried in a device-only branch shipped inert through two releases because nothing could execute it.

`planSchema()` takes `current` as a **parameter** rather than reading `kSchemaVersion`. With the constant pinned at 1, "treat legacy data as v1" and "stamp a blank device with current" produce the identical number, and a test written against the shipped constant cannot tell a correct implementation from one that conflates them. Every case that cares drives a synthetic `current` of 3.

**519 host tests green** (502 pre-existing + 17 new), WSL, `IDF_PATH` unset.

Every new test was verified to *discriminate* by breaking the production code and watching it fail — 17 independent mutations, **all 17 caught**, 10 of them by exactly one case, and every one of the 17 tests failed under at least one mutation:

| Mutation | Caught by |
| :--- | :--- |
| M1 legacy adopted at `current` not the v1 baseline | LegacyDeviceIsAdoptedAsV1AndNeverWiped, StampedZeroIsTreatedAsUnstamped |
| M2 blank device adopted at the v1 baseline | BlankDeviceIsBornCurrent |
| M3 a downgrade is allowed to stamp | NewerStampIsADowngradeThatWritesNothing, BlankDeviceIsBornCurrent |
| M4 unreadable store falls through to the normal path | UnreadableStoreIsNeverTreatedAsBlank |
| M5 an up-to-date device re-stamps every boot | CurrentDeviceWritesNothing, BlankDeviceIsBornCurrent |
| M6 a stored 0 honoured as a real version | StampedZeroIsTreatedAsUnstamped |
| M7 migrations never requested | OlderStampMigratesForward, LegacyDeviceIsAdoptedAsV1AndNeverWiped |
| M8 per-step stamp's return ignored | FailedStampAbortsTheChain |
| M9 a migration's return ignored | FailedStepStopsTheChainAtTheLastGoodVersion |
| M10 table search accepts any row at/past current | GapInTheTableFailsRatherThanSkipping |
| M11 shipped table reports a row | ShippedTableIsEmptyAtVersionOne |
| M12 `from == to` reported as failure | NothingToDoIsSuccess |
| M13 a backwards row is executed | BackwardsRowIsRefusedWithoutRunningIt |
| M14 downgrade outcome name changed | OutcomeNamesAreStable |
| M15 adopted legacy labelled a fresh install | LegacyDeviceIsAdoptedAsV1AndNeverWiped, LegacyDeviceOnBaselineFirmwareIsStampedWithoutMigrating, StampedZeroIsTreatedAsUnstamped |
| M16 per-step stamp never called | ChainRunsInOrderAndStampsAfterEveryStep, FailedStampAbortsTheChain, FailedStepStopsTheChainAtTheLastGoodVersion |
| M17 walker always short-circuits | 6 cases incl. EmptyTableCannotSatisfyAMigration |

**Hardware: untouched.** Host build and host tests only — the hardware pass is the human's. Nothing here has been flashed or run on a board, so the ESP-only glue in `ensureSchemaVersion()` (the `nvs_open`/probe sequence) is reviewed and reasoned about but not executed anywhere.

## Reviewer notes

- **Over the 500-line cap.** ~525 non-comment, non-blank lines (319 production + 206 test), against CONTRIBUTING §1.A. Flagging it rather than letting it be discovered. The raw diff is much larger again (~1080 lines) because of this repo's comment density; the header essay in `DeviceConfig.hpp` is where the downgrade argument and the decision table live, and I judged that worth the lines. Happy to split the docs/CHANGELOG out if preferred.
- `esp_log.h` is now included on the ESP side. It is in the same class as `nvs`/`esp_random`/`esp_partition` — a core component on every transport, not a WiFi dependency — and the standing rule against `esp_wifi.h` in this file is untouched, so the pure-Ethernet builds still link. The banner is emitted in `DeviceConfig` rather than in the four entry points so the one message an operator must see cannot drift between transports or be forgotten by a fifth entry point later.
- `CONTRIBUTING_FIRMWARE.md` §5 claimed **310** cases and is corrected, with an explanation of why a static grep reads 521 while 519 run on POSIX and 516 would run on Windows (two `#if !defined(_WIN32)` pairs in `DidMapping_test.cpp` / `TelephonyApiConfig_test.cpp`, holding 3+2 real cases against one `GTEST_SKIP` placeholder each).
- **Not done, deliberately:** the issue's point 4 (config export embedding `schema_ver`, import refusing newer and migrating older) is left for the export/import issue — there is no export format yet to embed it in. No dashboard field or `/api` surface for the outcome either.


GlomarGadaffi:
## Hardware-validated — including the case flagged as never-compiled and never-tested

The PR notes that the ESP-side glue *"has never been compiled"* and that the AdoptedLegacy path rests on reasoning rather than execution. Both gaps are now closed on the `.244` bench.

**Firmware builds clean** — `SIP_TRANSPORT=eth`, `PD_ETH_BOARD=elite`, IDF 6.0.2, zero errors. First compile of `probeNamespace` / `writeSchemaVersion` / `stampAfterStep` / the `nvs_get_u16`/`nvs_set_u16` calls and the `ESP_LOGE` banner.

**AdoptedLegacy, on a genuinely provisioned board.** The board carried a real admin credential, a dial-plan rule, a DID mapping and an SD-resident MoH clip, and had never seen a schema stamp. App-only flash to `0x20000` preserves NVS, so this is the exact scenario that would destroy a live device if the decision table were wrong:

```
W (624) cfgschema: adopting pre-versioning config as schema v1 (now v1); data preserved
```

And "data preserved" verified independently, not taken from the log line:

```
admin credential survived : YES
dial plan survived        : YES
moh clip survived (SD)    : YES
```

**Idempotent on the second boot** — the `UpToDate` branch, which is what keeps this from burning a flash write every power cycle:

```
I (623) cfgschema: schema up-to-date (v1)
```

**Ordering confirmed in situ.** The stamp lands at `624 ms`, before `esp_eth.netif` glue at `638 ms` and well before the SD mount at `713 ms` — i.e. before anything else opens a namespace READWRITE. That was the load-bearing claim about running ahead of `applyFlashSeed()`, and the boot ordering bears it out.

Not exercised here: `Migrated` and `Downgrade`, which have no second version to migrate to or from — as the PR says, rows for changes that have not happened would be untested flash-mutating code. `StoreUnavailable` likewise needs a corrupt store to provoke.

Issue #203, opened 2026-09-14
Blind transfer to a park orbit destroys the call: BYEs the caller, then 404s


GlomarGadaffi:
## Summary

**Blind-transferring a call to a park orbit destroys the call.** "Transfer to 701" is the classic way to park on any PBX; here the board accepts the REFER, tears the existing call down, and then reports `404 Not Found`. The caller is gone and nothing is parked.

Reproduced on hardware (`.244`), not inferred:

```
REFER: transfer the call to 701
  refer -> SIP/2.0 202 Accepted
  refer -> BYE sip:777@192.168.12.110:42374 SIP/2.0      <-- existing call torn down
  refer -> NOTIFY sip:1098@192.168.12.110:42374 SIP/2.0
     sipfrag: SIP/2.0 404 Not Found                       <-- transfer failed
```

## Root cause

Park orbits are reachable **two** ways and not a third:

| Path | Works? | Why |
| :-- | :-- | :-- |
| Direct dial `700` | yes | `onInvite` intercepts by number — `RequestsHandler.cpp:1476`, `_park.orbitIndex(destNumber)` |
| Dial-plan rule -> park | yes | `CallForker.cpp:374`, `DialActionType::ParkOrbit` |
| **Blind transfer / CFNA** | **no** | `CallForker::redirectInvite` starts with `findRegistered(target)` (`CallForker.cpp:150-154`) |

`findClient()` (`RequestsHandler.cpp:4713-4721`) scans `_clientPool`. A park orbit's stand-in peer is drawn from `_virtualPeerPool` via `allocVirtualPeer()`, which `findClient` never looks at — and it only exists *while a call is already parked there* anyway. So the lookup misses and `redirectInvite` returns false.

## Why the call dies rather than just failing

`onRefer` does the teardown BEFORE it knows whether the redirect will succeed (`RequestsHandler.cpp`, blind-transfer path):

1. BYE the other party
2. `endCall(callID, ...)`
3. `ok = targetClient.has_value() && _forker.redirectInvite(...)`
4. NOTIFY the transferor `200 OK` or `404 Not Found` based on (3)

Steps 1-2 are unconditional. By the time step 3 discovers the target is unreachable, the call it was supposed to move no longer exists. The transferor at least learns the truth here — the sipfrag really does say 404 — but the call is unrecoverable.

Note this interacts with #197: the same handler also moves the wrong party. These are separate defects in one function.

## Fix

[...]

Issue #128, opened 2026-09-01
Blind transfer: transferor leg is never torn down after REFER


GlomarGadaffi:
Found by the ported `.smoke/office_smoke.py` (scenario "Blind transfer": `B torn down=False, C received new INVITE=True`). Present on `main` as well as `feat/sdp-negotiation-invite-auth` — pre-existing drift from drawbridge (which passes 10/10), not fallout from the codec/auth work.

**Repro (host build, loopback):** A calls B; B answers; B sends REFER (blind) to C. C rings — that half works.

**Observed:** B (the transferor) is never torn down: no BYE reaches it after the REFER is accepted, so B sits in a dead call.

**Expected:** once the transfer target leg is created (and per RFC 5589 practice after the NOTIFY/sipfrag or immediately for the simple blind case), the transferor's leg gets a BYE — drawbridge's `onRefer` path does this.

**Where to look:** `RequestsHandler::onRefer` (blind-only in this repo); compare drawbridge's post-#178 behaviour ("blind-transfer silent drop" fix that never flowed back). `office_smoke.py` scenario `scenario_blind_transfer` is the regression test once fixed.

PR #211, opened 2026-09-14
Blind transfer moves the transferee, not the transferor (#197)

### LINKED ISSUE #197: Blind transfer moves the wrong party: the transferee is BYEd and the transferor is dialled to the target
## Summary

Blind transfer moves the **wrong party**. On a REFER, the PBX sends BYE to the party being transferred and re-INVITEs the **transferor** to the target. For the most common real-world shape — a receptionist transferring an inbound caller — that means **the customer is hung up on and the receptionist is dialled through to the target**, while the receptionist's phone reports success.

This is currently **asserted as intended by the test suite**, so this issue is a challenge to the semantics rather than a report of code not matching intent. That distinction is the point: the machinery is fine, the topology is inverted.

## What the code does

`RequestsHandler::onRefer`, `src/SIP/RequestsHandler.cpp:3842-3857`:

```cpp
if (auto original = getSession(callID); original.has_value())
{
    auto src  = original.value()->getSrc();
    auto dest = original.value()->getDest();
    auto other = (dest && dest->getNumber() != transferor->getNumber()) ? dest : src;
    if (other && other->getNumber() != transferor->getNumber())
    {
        auto bye = buildServerBye(other->getNumber(), other->getAddress(), callID, ...);
        if (bye) _outbox.emplace_back(other->getAddress(), std::move(bye));   // :3851
    }
}

endCall(callID, transferor->getNumber(), ..., "blind transfer");              // :3855

bool ok = targetClient.has_value()
        && _forker.redirectInvite(data, transferor, target);                  // :3857
```

`other` is *the party that is not the transferor* — i.e. exactly the party a transfer is supposed to move. It receives a BYE (`:3851`). The **transferor** is then re-INVITEd to the target (`:3857`).

## Why this is the wrong topology

RFC 3515 §2 / RFC 5359 §2.4: A and B are in a call. A sends `REFER` with `Refer-To: C` inside the A–B dialog. The intent is **B ends up talking to C**; A drops out. The referred request goes to the *referee*, not back to the referrer.

`tests/BlindTransfer_test.cpp:197-203` asserts the opposite and states it plainly:

> C — the transfer target — must get a fresh INVITE. … it now represents **A's continuation with C, not the torn-down A-B leg**.

And `:206-211` makes the impact explicit by naming the real-world case:

> The transferor can be either side of the original call — issue #128's own repro has the CALLEE transfer ("A calls B; B answers; B sends REFER"), **the more common real-world shape (a receptionist transferring an inbound call)**. `onRefer()`'s other/src/dest selection must **BYE the ORIGINAL CALLER** in that orientation.

Read that against what a receptionist expects when they press Transfer:

| | Expected | Actual |
| :-- | :-- | :-- |
| Inbound caller (A) | connected to C | **BYE — hung up on** |
| Receptionist (B, transferor) | drops out | **INVITEd to C** |
| Target (C) | talks to A | talks to B |

The caller who was meant to be transferred is disconnected, and the person who initiated the transfer gets the call instead. The transferor's phone shows success, because the `message/sipfrag` NOTIFY reports `200 OK` (`:3860`), so the lost call is invisible until the customer rings back.

## What is *not* wrong

The transferor-side REFER machinery is genuinely complete, including the part small stacks usually skip: the implicit-subscription NOTIFY with `Event: refer` and `Content-Type: message/sipfrag;version=2.0` is built at `:3894-3925`, citing §2.4.5. The BYE tag placement is correct and deliberately regression-tested (`BlindTransfer_test.cpp:188-195` — From carries the transferor's tag, To carries the other party's). The BYE itself was a real fix: without it the dropped party sat on a dead call forever (see the comment at `:3836`).

So this is not "transfer is broken everywhere" — it is one inverted decision about *which* leg survives.

## Secondary defects found alongside

Lower priority, but they belong to the same handler:

1. **The sipfrag is premature and can be false.** `:3857-3861` derives `"SIP/2.0 200 OK"` from `redirectInvite()`'s return value — which only means the INVITE was *queued* — and sends it with `Subscription-State: terminated`. §2.4.5 requires the referred request's **actual final response**. The transferor is told the transfer succeeded before the target has even been rung.
2. **No `100 Trying` NOTIFY** as the first notification (§2.4.4).
3. **`CSeq: 2 NOTIFY` is hardcoded** at `:3917`.
4. **REFER is never advertised** — there is no `Allow:` header anywhere in `src/`, and `Allow-Events` (`BlfSubscriptions.cpp:154`) lists only `dialog`, not `refer`.

## Suggested fix

Swap the topology: BYE (or re-INVITE away) the **transferor**, and send the referred INVITE to the **other party** so it lands on the target. `CallForker::redirectInvite(data, transferor, target)` at `:3857` takes the leg to move as its second argument — that is the single call site that decides this.

`BlindTransfer_test.cpp` will need updating in the same change, since it currently pins the inverted behaviour.

Worth confirming on the bench with the T29 before and after: place a call in, press Transfer, dial a third extension, and observe which party survives.

GlomarGadaffi:
Closes #197

A and B are talking, A REFERs to C. RFC 3515 §2 / RFC 5359 §2.4: **B** ends up talking to C, A drops out — the referred request goes to the referee, never back to the referrer. `onRefer()` did the reverse: it BYEd the party that is not the transferor (exactly the party a transfer exists to move) and re-INVITEd the transferor to the target. For a receptionist transferring an inbound caller, the customer was hung up on and the receptionist was dialled through, with the sipfrag reporting `200 OK` so the lost call was invisible until the customer rang back.

## Why it isn't a two-argument swap

Media is peer-to-peer here: the board relays offers and answers byte-for-byte and never carries audio. "B ends up talking to C" therefore requires B's and C's SDP to reach each other. `redirectInvite()` builds its INVITE **from the REFER**, which has no body at all, and reuses the A-B Call-ID — pointed at B it would produce a leg B has no dialog for and no media swap. A transfer that looks right in a trace and is silent on the handsets.

So a blind transfer is now the two-dialog B2BUA operation it has to be:

1. a **new leg toward C**, minted by the server impersonating B, carrying B's media (derived from which side of the dialog the transferor actually is — the receptionist case has them as the callee, so the SDP comes off the stored INVITE, not `getRemoteSdp()`, which is always "the callee's");
2. a **BYE to A** — the mirror image of the old one in every slot, tags included;
3. the **A-B session kept**, as B's half of a bridge. `endCall()` is not called: B stays in its own dialog, same Call-ID, same tags, no reconnect;
4. on C's 200 OK: **ACK it** (the server is that leg's UAC) and **re-INVITE B with C's SDP**, at CSeq 100 so `handleTransferOk()` ACKs B's answer instead of relaying it at an A who has gone.

Step 4 is ParkOrbit's retrieve swap split across time, because C has to be rung before it has an answer to swap in.

The #128 ordering comment is rewritten, not dropped: the collision it described was `redirectInvite()` reusing the Session under the A-B Call-ID, and the new leg has its own Call-ID and Session, so there is nothing left to order. `redirectInvite()` itself is untouched — CFB/CFNA still redirect the same party and still need it. **#203's property is preserved and widened**: the target lookup still gates every byte of teardown, and so now do the transferee lookup, its media, and both pools. Any miss declines with the call left up — and the call being protected is the one that used to be destroyed.

## Consequences of owning a leg the PBX did not own before

Each is a real wire failure and each has a test:

- **a non-2xx from the target is ours to ACK**, in the INVITE's own branch (§17.1.1.3) — otherwise it retransmits to Timer H. Claimed ahead of every session branch in `onBusy`/`onUnavailable`/`onFinalFailure`/`onReqTerminated`, because `onBusy`'s CFB lookup reads `getFromNumber()`, which on this leg is the **transferee**: a busy target would have forked the transferee's own forward-on-busy. The transferee is then released rather than left on a dialog with nobody on either end.
- **a 180 from the target is swallowed** — there is nobody to relay it to.
- **the transferor's own BYE no longer tears down the bridge.** Phones commonly hang up their leg the moment they see the NOTIFY (RFC 5589 §6.1) rather than waiting for ours, so that BYE lands on a dialog now re-purposed. The bridge-teardown branch relays only for the party still *on* the dialog. This covers attended transfer too.
- **a transferee who hangs up while the target still rings CANCELs** the outstanding INVITE (no To-tag exists yet to BYE).
- **a REFER retransmit does not re-run the transfer** (second commit). The A-B dialog now survives, so a second pass would invite the target twice and orphan the first leg — impossible under the old topology, where `endCall()` had erased the session. A survivor chaining a further transfer is declined 603 rather than guessed at: the src/dest derivation would pick the already-dropped transferor as the transferee, which is #197 again.

**Media direction is normalised to sendrecv** on both offers (`sipwire::sdpAsSendrecv`). Every Transfer softkey holds the call before it REFERs, so the last SDP captured for the transferee is routinely `a=recvonly`; relaying that verbatim completes the transfer with one-way audio — invisible in a trace, obvious on the handsets. Address and port are not touched: a phone that held with `c=0.0.0.0` needs its own re-INVITE to publish a live address, and inventing one would be a guess.

## Tests

`BlindTransfer_test.cpp` is rewritten rather than flipped — the topology its comments described is what changed. Per-party SDP (`sdpBodyFor`) replaces the single shared body, which could not tell a correct swap from a copy.

**509 host tests green** (502 before; 7 added, 2 rewritten in place). Every new and changed test was proven to discriminate — reverted production, observed the failure, restored:

| test | fails without the fix, at |
| :-- | :-- |
| `TransferorIsDroppedAndTheTransfereeIsMovedToTheTarget` | no BYE to the transferor |
| `ReceptionistTransferMovesTheInboundCallerNotTheReceptionist` | no BYE to the receptionist |
| `TargetAnswerIsAckedAndSwappedIntoTheTransfereesOwnDialog` | no ACK to the target |
| `HoldBeforeReferStillOffersTheTargetTwoWayMedia` | no `a=sendrecv` in the offer |
| `TransferorsOwnByeDoesNotTearDownTheCompletedTransfer` | `BYE sip:107@…` — the target hung up on |
| `TargetRefusalIsAckedAndTheTransfereeIsReleased` | no ACK to the 486 |
| `TransfereeHangingUpWhileTheTargetRingsCancelsTheLeg` | no CANCEL |
| `PostTransferByeFromTheTransfereeReachesTheTarget` | no BYE to the target |
| `ReferRetransmitIsAcceptedWithoutStartingASecondTransfer` | 2 INVITEs, 2 BYEs |
| `ReferSourceAuth.InDialogTransferFromARealLegStillSucceeds` | BYE at the wrong party, both directions |

Four sub-features were also proven independently, by disabling only that code and confirming **exactly one** test fails: the SDP normaliser, the stale-transferor BYE guard, the ringing-target CANCEL, and the failure-response hook. (Two HTTP tests flaked mid-session on a fixed-port collision with a stray background run of the binary — unrelated, green after `pkill`.)

## Not done — deliberately

- **The sipfrag is still premature** (#197 secondary items 1-2): it claims `200 OK` when the INVITE is queued rather than when the target answers, and there is still no `100 Trying` notification (§2.4.5 / §2.4.4). A separate defect with a different failure mode; folding it in would make this change about two things. It is arguably *more* visible now, since the answer genuinely arrives later.
- **REFER is still not advertised** in `Allow` / `Allow-Events` (items 3-4).
- **Post-transfer hold/resume from the transferee** relays toward the departed transferor — `onReinvite` works off src/dest, which still name the original pair. Attended transfer has the same gap today.
- **The session-timer sweep** on the original dialog would BYE the original pair, not the target.
- **No ring timeout on the target leg** — it relies on the target's phone sending a final response.
- **`onTrying` from the target** relays a 100 at the transferee under a foreign Call-ID. Harmless (a phone drops it), and the 180 case is handled; this one is four lines if wanted.
- **Chained transfer by the survivor** is declined, not implemented.

## Hardware

**Untouched.** Host build and host tests only; nothing has been flashed and the board has not been near this. The T29 bench pass the issue asks for — call in, press Transfer, dial a third extension, see which party survives and whether audio is two-way — is still owed, and it is the thing that would actually validate the media path. `tests/interop/interop.py`'s `blind_transfer` scenario asserted the old topology (it passed only when the *transferee* was torn down) and has been corrected in the third commit, also unrun, so the bench pass measures the right thing.


GlomarGadaffi:
## Hardware-validated, including the media path this PR rates "moderate, and explicitly not high"

Rebased onto `main` (which had moved by five PRs), built for `esp32s3`/`eth`/`elite`, flashed to the `.244` bench, and driven with three synthetic endpoints so no real handset was involved: **1096 transferor, 1095 transferee, 1094 target**.

```
registered 1096 (transferor), 1095 (transferee), 1094 (target)
  1096 -> 1095: INVITE delivered
  1095 answered; call is up

REFER: 1096 (transferor) transfers to 1094
  INVITE to 1094 (target)   : yes
  BYE to 1096 (transferor)  : True
  BYE to 1095 (transferee)  : False  <- must be False
  target was offered media at 192.168.12.110:59954 (B's RTP port is 59954)
  ...which is the TRANSFEREE's media: True
  re-INVITE to 1095 points at 192.168.12.110:33389 (C's RTP port is 33389) -> True

RTP: 1095 -> 1094 (peer-to-peer, 20 packets)
  1094 received 20/20 RTP packets from 1095

PASS: the TRANSFEREE was moved, not the transferor.
```

The two numbers that matter are the port comparisons, because they are what separates a real SDP swap from a body copied along unchanged:

- the target was offered **`59954`**, which is the *transferee's* RTP port — not the transferor's, and not a relayed copy of the original offer
- the transferee was re-INVITEd pointing at **`33389`**, the *target's* port

And then 20/20 RTP packets actually crossed between them, peer-to-peer, which is the thing host tests structurally cannot show.

**The topology inversion is gone.** `BYE to transferee: False` is the assertion that would have caught the original defect — a receptionist transferring a customer no longer hangs up on the customer.

## Rebase note

`main` gained `handleMohPreviewFailure()` at the same four final-failure handlers this PR adds `handleBlindXferFailure()` to, so all four conflicted. Resolved by keeping **both** — they match different Call-IDs and neither can mask the other, but dropping either sends that dialog back through `endHandle()`, whose lookup of a server-owned `From` matches no registered client, so the else branch answers a phone with a stray 404. That is the exact defect class both functions exist to fix.

583 host tests green on the rebased branch.

## Still open, from the PR's own list

Not re-litigating these — recording that hardware did not change them: premature sipfrag, missing `100 Trying`, REFER absent from `Allow`/`Allow-Events`, post-transfer hold/resume from the transferee still relaying toward the departed transferor, no ring timeout on the target leg, chained transfer declined.

Also unexercised here: a transferee that renegotiated its ports after the stored INVITE (stale stored SDP — the same caveat the attended splice already carries), and a phone that holds with the legacy `c=0.0.0.0` form.

Issue #222, opened 2026-09-14
DTMF factory reset (*PIN#999#1) doesn't wipe the SD CDR archive


GlomarGadaffi:
## Context

Issue #194 Stage 1 (SD-backed CDR archive) made the HTTP `/api/factory-reset`
path (`HttpServer::sendApiFactoryReset`) wipe the SD `/sdcard/cdr/` archive
alongside the existing NVS CDR ring wipe (`clearAllCallHistory()`), on the
explicit policy decision that the archive is treated as sensitive as
carrier credentials, same as the ring.

## The gap

The DTMF admin menu has its **own**, separate factory-reset trigger
(`*<PIN>#999#1`, handled in `DtmfFeatureCodes.cpp`). That path does not call
`HttpServer::sendApiFactoryReset` or `RequestsHandler::clearAllCallHistory()`
at all -- it runs `nvs_flash_erase()` followed by `esp_restart()` directly,
on the SIP thread.

`nvs_flash_erase()` wipes the entire NVS partition (more thoroughly than the
HTTP path's targeted per-namespace erase -- it takes out `cdrlog` along with
everything else), but it never touches the SD card's filesystem. So today:

- HTTP factory reset (`/api/factory-reset`): wipes NVS ring **and** SD archive.
- DTMF factory reset (`*<PIN>#999#1`): wipes NVS **but leaves the SD archive
  intact**.

Same "forget everything" action, two different real outcomes depending on
which door the operator used.

## Why it wasn't fixed in #194 Stage 1

`DtmfFeatureCodes.cpp` was out of that stage's scope boundary (the task was
scoped to `CdrArchive.{hpp,cpp}`, SNTP bring-up, and the minimal `endCall()`
hook). The DTMF factory-reset code also runs on the SIP thread, immediately
before `esp_restart()`, which is the same "no blocking file I/O under
`RequestsHandler::_mutex` / on the SIP thread" constraint `CdrArchive.hpp`'s
SD-write-discipline note describes for `endCall()` -- fixing this properly
needs either its own carefully-reasoned exception (the device is about to
reboot regardless, so a synchronous `opendir`/`unlink` sweep right before
`esp_restart()` may be acceptable, but that argument deserves review on its
own) or restructuring so both factory-reset entry points funnel through one
shared implementation.

## Scope

Either:
1. Add a direct (and justified) SD-wipe call to `DtmfFeatureCodes.cpp`'s
[...]

Issue #221, opened 2026-09-14
SD CDR archive: add direction, trunk/DID attribution, codec (needs Session-model plumbing)


GlomarGadaffi:
## Context

Issue #194 Stage 1 (SD-backed CDR archive) shipped an archive row with:
`timestamp, call_id, caller, callee, duration_sec, result, reason`.

The epic (#194) also asks for **direction**, **trunk/DID attribution**, and
**codec** on each archived row. These were deliberately left out of Stage 1
because none of them are cheaply available at the `endCall()` call site
(`RequestsHandler.cpp`) the way Call-ID and disposition `reason` were -- see
`src/SIP/CdrArchive.hpp`'s header comment for the exact reasoning.

## What's missing and why it's non-trivial

- **Direction** (inbound/outbound/internal): `Session` has no direction field.
  The closest cheap proxy -- checking whether `srcNumber`/`destNumber` matches
  a known virtual extension (440/555/888/777) -- only identifies *anchored*
  calls, not real inbound-vs-outbound trunk direction.
- **Trunk/DID attribution**: DID routing happens in `DidMapping`/`CallForker`
  well before `endCall()`, and nothing threads the matched DID (or "no DID,
  this was ext-to-ext") through to teardown. `Session` would need a new
  field, set at the point of routing, carried through the whole call
  lifecycle.
- **Codec**: no per-call codec is tracked anywhere today. Ordinary calls are
  peer-to-peer (the board never sees the codec choice at all -- see #194's
  "governing constraint" section); only the anchor/conference paths
  (`MediaBridge`/`MixBus`) touch media, and even those don't record the
  negotiated codec on the `Session`.

## Scope

Add whichever of these is decided to be worth it to `Session` (set at
routing/negotiation time) and thread it through to `CdrArchive::record()`'s
call site in `endCall()`. This is real call-routing/session-model surgery,
not a CDR-formatting change, hence its own issue rather than folding into
#194 Stage 1.

Issue #194, opened 2026-09-14
Epic: SD-backed CDR archive, Music on Hold, recording and voicemail


GlomarGadaffi:
## Goal

Use the microSD card on the T-ETH-ELITE for four consumers: **persistent CDRs** (distinct from the dashboard Call Log), **Music on Hold**, **call recording**, and **voicemail** — plus the two capabilities that make the ambitious versions of those work: **many conference legs** and **on-device decode for playback over the phone**.

The card is mounted and hardware-verified (#195): `/api/status` reports `{"present":true,"mounted":true,"capacityMb":480}` on `.244`. Nothing consumes it yet. This epic is the staged plan to change that.

> [!IMPORTANT]
> **#196 blocks Stages 2b, 3 and 4.** Those all require constructing or rewriting SDP, and pocket-dial has no SDP model — only a six-field line scanner that ignores `a=` attributes. Stage 1 (CDR archive) does not depend on it and can proceed in parallel.

## Relationship to existing issues

This epic was written before surveying the tracker. Several stages **already have issues, and in places those issues are better specified than the stage text below.** Where they conflict, the existing issue wins.

| Epic stage | Existing issue | Relationship |
| :-- | :-- | :-- |
| **2a** MoH for parked calls | **#162** | **#162 supersedes Stage 2a.** It specifies a *radio-station* model — one global clip cursor, every parked leg hearing the same stream, late joiners landing mid-song — explicitly **not** through MixBus, since park is one-way and there is no "self" to subtract. It also already flags that `RtpSender` is the wrong shape for this (4-instance pool with rx side and jitter buffer, too heavy for 10 parked legs), which corrects the "reuse `RtpSender::FrameProvider`" suggestion below. |
| **2a/4** SDP rewriting | **#196**, and #162 independently | #162 already named the blocker: `SipSdpMessage::rewriteMediaTarget(serverIp, port)` does not exist. That is the same gap #196 covers. |
| **5** PIE vector kernels | **#171**, **#145** | #171 is kernel 6b (int32 sum-once-subtract) and records that it needs ops not verified to assemble, plus the QACC `ee.vmulas.s16.qacc` alternative. #145 is the one-sided `#ifdef` — `mix_kernels_scalar.cpp` guards with `#ifndef POCKETDIAL_MIXBUS_PIE` but the `.S` has no guard at all, so the opt-in only has one half. |
[...]

PR #223, opened 2026-09-14
feat(cdr): SD CDR archive, Stage 1 of #194

GlomarGadaffi:
## What this is

Issue #194 Stage 1 only: an append-only SD-card CDR archive
(`/sdcard/cdr/YYYY-MM-DD.csv`), separate from and additive to the existing
32-slot NVS `CdrRing` used by the live dashboard Call Log. No other stage
of #194 (MoH, voicemail, recording) is touched.

**Part of #194** (the epic stays open -- Stages 2-5 are unaddressed and
separately scoped; this PR does **not** close it).

## Prerequisite 1a (wall clock) -- already landed, not duplicated

Per the task instructions, I checked first: `src/Helpers/TimeSync.hpp`/`.cpp`
(SNTP bring-up, RFC 3339 formatting) and its call from
`main/esp_main_eth.cpp`'s Ethernet `GOT_IP` handler were already on this
branch before this work started (commit `a6820c8`, landed the same day this
task began). This PR reuses `timesync::epochSeconds()` as-is.

## Prerequisite 1b (SD write discipline) -- enforced

`CdrArchive` never touches a file on the SIP thread or under
`RequestsHandler::_mutex`:
- `endCall()`'s only new line is `cdrarchive::record(rec, callID, reason)` --
  non-blocking, formats a fixed-size line (no heap) and pushes it into an
  in-process queue.
- A dedicated low-priority (`priority 1`, matching `log_drain`) FreeRTOS task,
  spawned from `cdrarchive::init()` (called right after `sd_mount()` in
  `app_main`), drains that queue and does the actual `fopen`/`fwrite`/`fclose`
  against `/sdcard/cdr/<date>.csv`.
- The in-process queue (`WriterQueue`) is touched from three tasks on real
  hardware (SIP thread push, writer task pop, HTTP task clear on factory
  reset) and is mutex-guarded accordingly -- a leaf lock, nothing under it
  blocks.

## Design

- **New module**: `src/SIP/CdrArchive.{hpp,cpp}`. Pure CSV formatter
  (`formatLine`) + queue/drain mechanics (`WriterQueue`/`drainAll`) are
  platform-neutral and fully host-tested via an injectable `Sink`. The real
  `FatFsSink` + writer task are gated on `PD_ETH_HAS_SD` specifically (not
  bare `ESP_PLATFORM`) -- the WiFi and LAN8720 transports build for boards
  with no SD slot and don't even declare `pd_sd_mounted()`, so a bare
  `ESP_PLATFORM` guard would be a link error those builds can't catch from a
  host-only session. Every build without `PD_ETH_HAS_SD` (host, wifi,
  lan8720, or eth/waveshare) degrades silently: no `Sink` is ever installed,
  so `record()`/`wipeAll()` are harmless no-ops.
- **`endCall()` hook** (`RequestsHandler.cpp`, the `_cdr.record(...)` call
  site): `CdrRing::record()` now *returns* the `CallDetailRecord` it just
  wrote (previously `void`), so the archive reuses the already-derived
  `startMs`/`durationSec`/`result` (including the `Connected`/`Held`/`Bye` ->
  `Answered` disposition logic) instead of re-deriving them, and gets the
  Call-ID and disposition `reason` the ring itself doesn't carry. `CdrRing`'s
  fields, NVS format, and dashboard view are otherwise **completely
  unchanged** -- "SD is the archive, not a replacement."
- **Row shape**: `timestamp,call_id,caller,callee,duration_sec,result,reason`.
  Timestamp is the call's **start** (derived from `nowEpochSeconds` minus how
  long ago `rec.startMs`, a steady-clock reading, was), not the archive
  moment. `record()` drops the entry entirely if the clock has never synced
  (`epochSeconds() == 0`) rather than filing it under a fabricated date --
  the only calls affected are ones that end in the boot-to-first-SNTP-reply
  window, and they're still in the NVS ring exactly as before this PR.
- **CSV escaping**: RFC 4180, applied to `call_id` and `reason` (not
  charset-constrained) and, defensively, to `caller`/`callee` too -- see the
  `isValidAor` finding below.

## Factory-reset decision (explicit, per the task)

**`/api/factory-reset` now wipes the SD archive too**, same policy as the
existing NVS ring wipe (`clearAllCallHistory()`). This codebase already
treats call history as sensitive as carrier credentials -- that's the whole
reason `clearAllCallHistory()` is its own explicit step rather than falling
out of `DeviceConfig::clearAll()`. Leaving a full, dated, plaintext archive
on the card after a reset that just erased the live ring would be the more
surprising and worse outcome of the two. **If this default is wrong for how
pocket-dial is actually deployed (e.g. the SD archive is meant to be the
durable compliance record), it's a one-line reversal** at the
`cdrarchive::wipeAll()` call site in `HttpServer::sendApiFactoryReset` --
flag it in review.

Implementation note: `wipeAll()` is called directly from
`sendApiFactoryReset` (HTTP task, no lock held), **not** threaded through
`clearAllCallHistory()` -- that method holds `RequestsHandler::_mutex`, and
directory I/O must never run under it.

**Known gap, filed separately, not fixed here**: the DTMF admin menu's own
factory-reset path (`*<PIN>#999#1`, `DtmfFeatureCodes.cpp`) runs
`nvs_flash_erase()` + `esp_restart()` directly on the SIP thread and never
reaches this code at all -- so a DTMF-triggered reset currently leaves the SD
archive intact while the HTTP-triggered one (this PR) wipes both. Filed as
#222; out of this stage's scope (`DtmfFeatureCodes.cpp` wasn't in the
authorized touch list, and fixing it properly needs its own reasoning about
synchronous I/O immediately before a reboot).

## Other defects fixed while in here

- **`CallDetailRecord.hpp`'s comment was false**: it claimed caller/callee
  "are bounded by the registrar's `isValidAor()` validation... so they
  cannot grow without limit." `isValidAor()` (`RequestsHandler.cpp`) is
  charset-only with **no length check at all**, and the anchor/trunk paths
  (`ev.participantId`, `Session::getAnchorParticipantId()`) bypass it
  entirely, sourcing values from the anchor's own JSON. Fixed the comment
  to say so precisely, and added a real 64-char bound to `isValidAor()`
  itself (small, isolated change -- charset check unchanged, new length
  guard is one line) since an unbounded AOR heap-allocates inside every
  fixed-footprint `CdrRing` slot. `CdrArchive` additionally truncates
  caller/callee/call-id defensively on its own, since anchor-sourced values
  still bypass `isValidAor()` regardless of this bound.
- **The five-field `CallDetailRecord` is thin for an archive** (no
  direction, trunk/DID attribution, codec) -- deliberately not extended
  here (would touch the ring's fixed-footprint slot type, its NVS blob
  shape, and the dashboard JSON all at once, and none of those three fields
  are cheaply available at `endCall()` without real call-routing/session
  plumbing). Filed as #221.

## Test plan

- [x] Host build + full suite green via WSL (`cmake` + `ctest`), 606 tests
      across 91 suites, 0 failures, including 26 new/changed tests:
  - `tests/CdrArchive_test.cpp` (22 tests): `formatLine()`'s wall-clock
    derivation (including both clamp directions), CSV field
    order/escaping/truncation, the fixed-buffer-never-overflows worst case,
    `WriterQueue` push/pop FIFO order/drop-when-full/clear/zero-capacity,
    `drainAll()`'s queue-to-sink decoupling, and the singleton
    `record()`/`wipeAll()` entry points' host-reachable no-op/wipe contract.
  - `tests/Pbx_test.cpp` (+1 test, 1 updated): the new 64-char `isValidAor`
    bound, mirrored in the existing charset-mirror test helper.
- [x] Confirmed (before writing any code) that SNTP bring-up was already on
      this branch, per the task's explicit instruction to check first.
- [ ] **NOT compiled**: the actual ESP-IDF firmware (`idf.py`) for
      `SIP_TRANSPORT=eth PD_ETH_BOARD=elite` -- this was a host-test-only
      task with no ESP-IDF toolchain available. The `main/CMakeLists.txt`
      source-list addition and the `PD_ETH_HAS_SD` code path are correct by
      inspection (mirrors the existing `CdrRing.cpp`/`sd_mount()` pattern
      exactly) but unverified by an actual cross-compile.
- [ ] **NOT hardware-tested** (per task rules -- no hardware was touched):
      the writer task's 6144-byte stack is a reasoned estimate, not measured
      with `uxTaskGetStackHighWaterMark()`; the real `FatFsSink`'s
      `opendir`/`readdir`/`unlink`/`mkdir`/`fopen` calls against FatFs have
      never run; nothing was flashed to any board.

## Followups filed

- #221 -- direction/trunk-DID-attribution/codec fields (needs Session-model
  plumbing, out of scope here).
- #222 -- DTMF factory reset doesn't reach the SD archive wipe.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_013LMLqvEZy1Pv5T4pSnPp3j


GlomarGadaffi:
Pushed a fix for both findings from the review pass (commit 8223d46):

**TOCTOU race (blocking)**: added `drainWipeMutex()`, held across the writer task's whole drain-everything-pending loop and across `wipeAll()`'s whole clear+wipe sequence, so the two are mutually exclusive as complete operations rather than only individually thread-safe. This is a structural fix (provably race-free by mutual exclusion, not something that needs a timing-dependent test to demonstrate) — the free `drainAll(WriterQueue&, Sink&)` function used by host tests is unchanged.

**Unconditional allocation on no-SD boards (should-fix)**: `wipeAll()` now checks the sink first and returns before touching `queue()`, mirroring `record()`'s existing correct order — so a factory reset on a board with no SD archive installed never forces the writer queue's ~19.5KB backing vector into existence at runtime.

606/606 host tests green under WSL, no regression (count unchanged from the original PR).

Issue #185, opened 2026-09-14
Reliability: task watchdog + heap/stack high-water mark monitoring


GlomarGadaffi:
> Ported from drawbridge #127. Verified gap here: no `esp_task_wdt`/`uxTaskGetStackHighWaterMark` in `src/`+`main/`; no explicit `TASK_WDT` config.

A wedged task (HTTP worker on a slow client, SIP tick on a lock, `conf_mix_tick` with its 3072-byte stack) goes undetected until the unit is unresponsive.

1. Subscribe `udp_receiver_task`, `sip_server_task`, `rtp_media_tx/rx`, `conf_mix_tick` to the IDF Task WDT; stall → controlled reset with a logged cause.
2. Expose `uxTaskGetStackHighWaterMark()` per named task as `stackHwm_<task>` in `/api/status`.
3. Expose `esp_get_minimum_free_heap_size()` and `heap_caps_get_minimum_free_size(MALLOC_CAP_SPIRAM)` alongside `freeHeap`.
4. Make the host-test `heap_debug` config record these in the interop harness.

Note from the mixer survey: `MixBus::tick()` puts ~2.9 KB of locals on a 3072-byte stack — the HWM number will tell us whether that's already marginal.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01JJSmyBFe3aesr4id9pEMm7

PR #225, opened 2026-09-14
Reliability: Task WDT for sip_server_task + heap/stack telemetry on /api/status (#185)

GlomarGadaffi:
## Summary

Addresses #185 (partial — see #235 for the remaining task-watchdog subscriptions and interop recording). Watchdog subscription and heap/stack telemetry are done for sip_server_task; udp_receiver_task, rtp_media_tx, rtp_media_rx and conf_mix_tick still need their own watchdog subscriptions and are tracked in #235.

1. **Task Watchdog**: `sip_server_task` is subscribed to the IDF Task Watchdog
   Timer in all three in-scope entry points (`main/esp_main.cpp`,
   `main/esp_main_eth.cpp`, `main/esp_main_eth_lan8720.cpp`) —
   `esp_task_wdt_add(NULL)` once, `esp_task_wdt_reset()` every ~1s tick.
   `sdkconfig.defaults` gets `CONFIG_ESP_TASK_WDT_PANIC=y` — without it a
   timeout only logs a warning and the stall is left exactly as wedged, which
   is not the "controlled reset with a logged cause" the issue asks for.
   Verified end-to-end against the actual ESP-IDF v6.0.2 source in this box's
   toolchain, not assumed:
   - `esp_task_wdt_reset()` only ever resets the **calling** task's own
     subscription (`components/esp_system/include/esp_task_wdt.h`) — there is
     no "reset on behalf of task X" call.
   - `esp_reset_reason_set_hint(ESP_RST_TASK_WDT)` is called before the panic
     abort (`task_wdt.c:380`), so `resetReason":"TASK_WDT"` on the *next* boot
     is a correct claim, not `PANIC`.
   - All three generated `sdkconfig`s (wifi/eth/lan8720) confirmed:
     `CONFIG_ESP_TASK_WDT_PANIC=y`, `CONFIG_ESP_SYSTEM_PANIC_PRINT_REBOOT=y`
     (not HALT/GDBSTUB), `CHECK_IDLE_TASK_CPU0/1=y`, `TIMEOUT_S=5`.

2. **`udp_receiver_task`, `rtp_media_tx`, `rtp_media_rx`, `conf_mix_tick` are
   NOT subscribed** to the watchdog in this PR — their loops live in
   `UdpServer.cpp` / `RtpSender.cpp` / `RtpReceiver.cpp` / `ConferenceRoom.cpp`,
   all outside this change's declared file scope (`main/esp_main*.cpp` +
   `sendApiStatus` only). Since `esp_task_wdt_reset()` can only be called by
   the subscribed task itself, subscribing them from outside their own file
   would guarantee a spurious reset under normal operation rather than add
   safety, so I left them unsubscribed instead of doing that. Exact follow-up
   locations (documented in `docs/API.md`'s new Task-Watchdog note too):

   | Task | File | Where `esp_task_wdt_add`/`_reset` would go |
   | :-- | :-- | :-- |
   | `udp_receiver_task` | `src/Helpers/UdpServer.cpp` | `add` at the top of `receiveLoop()` (~L163); `reset` once per `while(_keepRunning)` iteration; `delete` after the loop exits, before the task signals `_receiverExited` |
   | `rtp_media_tx` | `src/SIP/RtpSender.cpp` | `add` at the top of the task trampoline (~L303); `reset` once per `while(!_stopRequested...)` iteration (~L342) |
   | `rtp_media_rx` | `src/SIP/RtpReceiver.cpp` | `add` in `taskTrampoline`/`runLoop()` (~L353/363); `reset` once per `while(!_stopRequested...)` iteration (~L388) |
   | `conf_mix_tick` | `src/SIP/ConferenceRoom.cpp` | `add` at the top of `runDriver()`; `reset` once per tick (~L180) |

   `rtp_media_tx`/`rx` and `conf_mix_tick` are ephemeral (created per active
   call/conference), so their follow-up also needs `esp_task_wdt_delete(NULL)`
   before task exit or the TWDT keeps a stale entry.

3. **`GET /api/status`** (`sendApiStatus`, `src/Helpers/HttpServer.cpp`) gains,
   purely additively — the existing hunk above the new lines is byte-for-byte
   unchanged:
   - `freeHeap`, `minFreeHeap`, `minFreeHeapSpiram` — the issue text said
     "alongside the existing `freeHeap` field", but no such field existed;
     `sendApiStatus`'s own pre-existing comment said so explicitly
     ("`heap_free_bytes` / `psram_free_bytes` have no counterpart in this
     server's status route"). All three are new here, not additions to an
     existing field.
   - `resetReason` — `esp_reset_reason()` decoded to a short label.
   - `stackHwm_sip_server_task`, `stackHwm_udp_receiver_task`,
     `stackHwm_rtp_media_tx`, `stackHwm_rtp_media_rx`, `stackHwm_conf_mix_tick`
     — `uxTaskGetStackHighWaterMark()`, looked up by FreeRTOS task **name**
     via `xTaskGetHandle()` rather than by threading a `TaskHandle_t` out of
     `SipServer`/`UdpServer`/`RtpSender`/`RtpReceiver`/`ConferenceRoom` into
     `HttpServer.cpp` — so none of those classes needed a new getter, and item
     2's four out-of-scope files stay completely untouched by this endpoint
     too. `null` (not `0`) when a task isn't currently running: `0` is the
     real near-overflow alarm this field exists to catch, and collapsing
     "not running" into the same value would hide it. `udp_receiver_task`'s
     name is queried pre-truncated (`"udp_receiver_ta"`) because
     `xTaskGetHandle()` asserts `strlen(name) < configMAX_TASK_NAME_LEN` (16
     on this project's sdkconfig) and FreeRTOS silently truncates the stored
     name the same way at task-creation time — verified against
     `FreeRTOS-Kernel-SMP/tasks.c`, not assumed; querying the untruncated
     18-char name would trip that assert and abort the board.
   - Host build emits the identical key set as `0`/`null` so
     `tests/interop/interop.py`'s JSON parsing never has to special-case
     platform.

4. **Item 5 (host-test `heap_debug` interop recording) is a followup, not
   done here.** `tests/interop/` runs the **host build** of the PBX binary
   (`subprocess.Popen`), and every one of these new fields is ESP-only
   (`#if defined(ESP_PLATFORM)`) — on the host build they're always `0`/`null`
   by construction. Wiring the interop harness to "record" them would only
   ever capture host zeros; there is nothing ESP-specific to observe without
   real hardware, so I didn't add a no-op recording step for it.

## Known gaps / trade-offs, disclosed rather than hidden

- **One file outside the stated scope**: `sdkconfig.defaults` got
  `CONFIG_ESP_TASK_WDT_PANIC=y`. Without it, item 1's subscription would only
  ever print a warning on timeout — the "controlled reset" half of the issue
  would not exist. I judged this a necessary, minimal, disclosed exception
  rather than shipping a watchdog that can't actually reset anything.
- **`docs/API.md` and `docs/THREAT_MODEL.md` also touched** (not in the file
  list either) — `docs/API.md`'s `/api/status` section explicitly says "this
  is the full field set, not an abridgement", so leaving it stale would be
  exactly the kind of claim #216's audit was about. `THREAT_MODEL.md` §4 E-2
  is the route's gating justification; I added one paragraph classing the new
  fields as unlabelled aggregates (the same class already argued ungated for
  `/metrics`), not the roster/session/dial-plan class E-2(2) is about — no
  reassessment of `/api/status`'s gating needed.
- **`main/esp_main_display.cpp` is untouched** (correctly out of scope per the
  file list, and excluded from the ESP-only build-verification list too) —
  but `sdkconfig.defaults`'s `CONFIG_ESP_TASK_WDT_PANIC=y` still applies to
  that build. Its `sip_server_task` has no TWDT subscription and no
  reset-reason log, so the display variant now gets idle-task
  panic-on-starve with none of this PR's matching benefit. Same three-part
  edit (subscribe + reset + boot-reset-reason log) is a clean, isolated
  followup for that file.
- **`xTaskGetHandle()` suspends the scheduler** (`vTaskSuspendAll()`) for a
  short walk of the task list, once per lookup, five times per
  `/api/status` poll. This is the standard IDF way to do a name lookup and
  the walk is short/bounded (no allocation, no lock), but on a
  real-time-audio board it's worth naming: the dashboard polls this route
  every ~2s in normal use, not per-packet.
- **`tick()` has ~4s of slack** before the 5s TWDT timeout at today's 1s loop
  cadence. If `tick()` ever grows a blocking I/O call, this margin — and the
  hardcoded `TIMEOUT_S=5` — would need revisiting together.

## Test plan

- [x] Host build + full suite via WSL (`cmake -B build-wsl && cmake --build
      build-wsl && ctest --test-dir build-wsl/tests`): **583/583 gtest cases
      passed** (measured via the binary's own `[==========] 583 tests from 88
      test suites ran... [ PASSED ] 583 tests.`), 0 failed.
- [x] ESP-IDF v6.0.2 build, all three variants, **all green**:
  - `idf.py -D SIP_TRANSPORT=wifi set-target esp32s3 build` → `SipServer.elf`
    linked, 0 `FAILED` lines.
  - `idf.py -D SIP_TRANSPORT=eth set-target esp32s3 build` (PD_ETH_BOARD=elite
    default) → "Project build complete", 0 `FAILED` lines.
  - `idf.py -D SIP_TRANSPORT=lan8720 set-target esp32 build` → 0 `FAILED`
    lines, app + bootloader images generated.
  - Confirmed by grep that every edited file (`esp_main.cpp`,
    `esp_main_eth.cpp`, `esp_main_eth_lan8720.cpp`, `HttpServer.cpp`) actually
    compiled as its own object in each relevant build log, not merely that
    the overall build exited 0.
  - Confirmed by grep on each variant's generated `sdkconfig` that
    `CONFIG_ESP_TASK_WDT_PANIC=y` and the related TWDT/panic settings landed
    as intended.
- [ ] **Not hardware-verified** (no board touched, per this task's rules): a
      real TWDT trip actually resetting a board, `xTaskGetHandle`'s
      truncated-name match against a live FreeRTOS task table, and the actual
      numeric HWM/heap values under load are all source-verified and
      build-verified only.
- [ ] Item 5 (interop `heap_debug` recording) intentionally not done — see
      "Known gaps" above.

## Environment notes for whoever touches this box next

- The Bash tool refuses `wsl.exe` invocations from a worktree-isolated agent
  (worried it can't verify the command stays inside the worktree), even for
  read-only commands — the PowerShell tool does not have this restriction and
  was used for every `wsl.exe`/`idf.py` call here.
- The shared `C:\Users\desmo\esp\esp-idf` checkout has **CRLF line endings**
  on its shell scripts (`export.sh`) and on at least one Python entry point's
  shebang line, so `. export.sh` from WSL bash fails
  (`$'\r': command not found`, then `env: 'python\r': No such file`).
  `export.ps1` run natively via the PowerShell tool is the working path on
  this box for the ESP-IDF toolchain specifically (WSL remains correct for
  the CMake/ctest host suite, per this task's own instructions).
- Parallel **first-configure** `idf.py` runs across sibling project
  directories race on the shared
  `%LOCALAPPDATA%\Espressif\ComponentManager\Cache\` directory
  (`WinError 145: directory not empty` mid-`rmtree`) — stagger first builds,
  or pre-populate `managed_components/`+`dependencies.lock` once before
  parallelizing.
- This worktree's path (`...\Glomar's Github\...`) breaks ESP-IDF builds two
  ways: the literal apostrophe, and, independently, `CMAKE_OBJECT_PATH_MAX`
  (250 chars) — object paths for `mbedtls/tf-psa-crypto` land at 204-227 chars
  under a long build dir and the `.obj.d` dependency file write fails
  (`fatal error: opening dependency file ...: No such file or directory`).
  Fix used here: keep the **source** wherever (a short WSL-native path solves
  the apostrophe; a Windows scratch copy also works), but always pass a
  **short** `-B <builddir>` (e.g. `C:\pdbuild\<variant>`) — only the build
  directory's path length matters for this failure mode.
- One transient `-1073740022` (`STATUS_DLL_INIT_FAILED`) build failure was
  observed generating `esp_system/ld/sections.ld` while 3 heavy `idf.py`
  builds ran concurrently on this box; a solo retry of the same variant
  passed cleanly. Treat a lone failure at that exact step as resource
  contention, not a real break, and retry before concluding otherwise.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_013LMLqvEZy1Pv5T4pSnPp3j


GlomarGadaffi:
The watchdog-repositioning fix (commit b514764) duplicated the block instead of moving it: \`esp_main_eth.cpp\` and \`esp_main_eth_lan8720.cpp\` both ended up with two \`esp_err_t wdtErr = esp_task_wdt_add(NULL);\` declarations in the same function scope (one before \`new SipServer(...)\`, left over from the original code, plus a new one immediately before the main loop) — that's a redeclaration and won't compile. \`esp_main.cpp\`'s fix was correct (single call, properly moved).

Neither the host build nor the host test suite catches this class of bug — these three files are ESP-only and never enter the \`sip_parser_tests\` target, so a green host suite was never evidence this compiled. This is the same blind spot PD-Opus hit tonight with a link-only failure in #233's DTMF work: `main`-level files need either a real ESP-IDF build or a careful read, a green host suite alone proves nothing about them.

Pushed a fix (commit cbf8259): removed the stale pre-construction block in both files, keeping only the correctly-positioned one. Confirmed by inspection — single `wdtErr` declaration in each file now, brace-balanced, and diffed clean (pure 15-line deletions, nothing else touched). Host suite re-run for completeness: 583/583, unaffected as expected since these files aren't in that target.

PR #237, opened 2026-09-15
E.164 normalization: a DID now matches however it is written (#165)

### LINKED ISSUE #165: Dial plan has no outbound-trunk routing action, and no E.164 normalization anywhere

GlomarGadaffi:
Closes #165.

## Most of #165 already shipped

Issue #165 is titled "Dial plan has no outbound-trunk routing action, **and** no E.164 normalization anywhere". The first half is done and has been for a while — `d47c131` added `DialActionType::Trunk` with `stripDigits`/`target`, and it is wired end to end:

`CallForker.cpp:398` `pbx::applyTrunkTransform()` → `PbxEnv::routeTrunkCall()` → `RequestsHandler::originateAnchorCall()`

The dialed digits do reach the anchor, and the double-answer hazard is already handled — `routeTrunkCall()` returns false *only* when no anchor is connected and deliberately sends nothing, so `CallForker` falls through to its shared 404 rather than answering the same INVITE twice. `.244`'s live `/api/status` shows a working rule of this shape.

This PR is the other half, which genuinely did not exist. Six files mention E.164 in comments; nothing implemented it.

## The defect

`DidMapping::findIndex()` compared DIDs with a raw `std::string ==`. So this never matched:

| configured by the operator | reported by the carrier | matched before |
|---|---|---|
| `(202) 555-0123` | `+12025550123` | ❌ |
| `+12025550123` | `2025550123` | ❌ |
| `202-555-0123` | `202.555.0123` | ❌ |

**And it fails silently.** `routeInboundAnchorCall()` reads "no mapping" as "fall back to ring-all", so the call still rings phones — just not the one the operator configured. No error, no log, nothing that looks like a bug. It surfaces months later as "why did the wrong person answer".

## What was added

**`pbx::e164Normalize()`** is a formatter, not a resolver:

- strips the ways humans decorate a number — space, tab, `-`, `.`, `(`, `)`, `/`
- keeps a leading `+`, the only thing that makes a country code unambiguous
- enforces E.164's 15-digit ceiling (ITU-T E.164 §6.2.1) by **refusing, not truncating** — a truncated number is a different, possibly valid, line
- refuses anything that is not a number. `*` and `#` are *real dialable symbols* in this PBX (`*8` group pickup, `**<ext>` directed pickup), so a feature code must never be stripped down into something that routes
- **never invents a country code.** `5551234567` normalizes to `5551234567`, never to `+15551234567`

`00` is deliberately *not* treated as an international prefix: in NANP the IDD prefix is `011` and `00` is an operator code, so that rule would be wrong in exactly the numbering plan this firmware targets — and carrier-reported DIDs never arrive that way.

**`pbx::e164SameNumber()`** is where the judgement lives. Two renderings are the same line when they differ only by an international prefix, bounded on both ends:

- the shorter side must reach `POCKETDIAL_MIN_PSTN_AOR_DIGITS` (7 — the same threshold that already separates "PSTN-shaped" from "internal extension"), so a 3-digit extension can never tail-match its way into a DID
- they may differ by at most **3** digits, the actual ITU bound on a country code — not an arbitrary number
- two `+`-marked numbers must match **exactly**, since each has already declared its own country

### The boundary, stated rather than hidden

One false positive is knowingly accepted: `+442025550123` and `2025550123` compare equal. A bare national number carries no country, so relating it to an E.164 number *requires* assuming one.

Refusing the relaxation entirely would fail the common case — an operator typing the 10-digit DID they know — every single time, to avoid a collision that needs a deliberate coincidence inside a table of at most `POCKETDIAL_MAX_DID_MAPPINGS` (8) entries. `E164.hpp` documents it, and **a test pins it** so it cannot change unnoticed. If a deployment ever needs strictness, the fix is a configured default country code, not a tighter heuristic.

## Backwards compatibility

`DidMapping::sameDid()` tries exact string equality **first and unconditionally**. `e164SameNumber()` returns false for anything it cannot normalize, so without that line a hand-edited non-numeric DID would become an unremovable row. E.164 equivalence is added *on top of* the old behaviour, never in place of it.

Two consequences, both documented in `DidMapping.hpp`:

- `setMapping()` with a different rendering of an existing DID **updates** that entry rather than adding a second one. Deliberate: two rows matching one inbound call would make routing depend on table order, which an operator can neither see nor control.
- `removeMapping()` likewise accepts any rendering.

## Testing

24 new tests in `tests/E164_test.cpp`. Roughly half assert what must **not** match, because a false positive here routes a stranger's call to the wrong desk.

**Every guard was mutation-tested.** All 8 mutations were caught by a failing test:

| mutation | caught by |
|---|---|
| `sameDid` exact-match fallback removed | `ANonNumericDidRemainsFindableAndRemovable` |
| `findIndex` reverted to raw `==` | all 8 `DidE164Test` cases |
| both-international guard removed | `TwoFullyQualifiedInternationalNumbersMustMatchExactly` |
| significant-digit floor removed | `AShortExtensionCannotTailMatchItsWayIntoADid` |
| country-code length bound removed | `APrefixLongerThanACountryCodeIsNotACountryCode` |
| 15-digit ceiling → truncation | `EnforcesTheITUFifteenDigitCeilingByRefusingNotTruncating` |
| `*`/`#` treated as separators | `RefusesFeatureCodesAndVanityNumbersRatherThanManglingThem` |
| interior `+` silently dropped | `RefusesAPlusThatIsNotTheFirstSignificantCharacter` |

The suite was rebuilt after restoring each mutation, so none of this is a stale binary reporting green.

- Full host suite: **820/820** under WSL
- Every number in the tests is from NANP's reserved fictitious range (555-0100..555-0199)

## What this does NOT prove

Nothing here is exercised against a real carrier. Only the single configured route DN can currently reach `extensionForDid()` (there is no per-call dialed-DID field yet — see `DidMapping.hpp`'s BOUNDARY note), so per-DID matching is proven at the data-model level, **not** end to end through an inbound PSTN call.

## Hardware

Untouched. Compile-only via the host suite; no flash.


GlomarGadaffi:
Vetted independently (same GitHub account as the PR author, so this is a comment not a formal approval — verdict: **ship it**).

## What I checked
- Read `E164.cpp`/`.hpp` and `DidMapping.cpp`/`.hpp` in full, hand-traced `e164SameNumber()` against every documented case: equal-length/formatting-only, the `+`-marker-only case, national-vs-international with a country code present/absent, the both-`+` refusal, the country-code-length bound, the `POCKETDIAL_MIN_PSTN_AOR_DIGITS` floor, and the one knowingly-accepted false positive (`+442025550123` vs `5551234567`). All correct by hand.
- Built the PR branch from scratch in an isolated worktree (not trusting the PR's own claimed numbers) and ran the full suite: **820/820**, matches exactly.
- Ran just the 24 new tests (16 `E164*` + 8 `DidE164Test`) in isolation — all pass, none vacuous, and `MatchingSurvivesAStoreRoundTrip` actually exercised the persistence round-trip (ran on WSL/POSIX, correctly self-skips on `_WIN32`).
- Spot-checked one of the claimed mutation-testing results myself rather than trusting the table: mutated out the both-`+`-refuses guard (`if (plusA && plusB) return false;` → `if (false && ...)`), rebuilt, reran `TwoFullyQualifiedInternationalNumbersMustMatchExactly` — it failed exactly as claimed (`+12025550123` vs `+2025550123` incorrectly returned `true`). Reverted. This is the safety-critical guard (merging two different countries' numbers), and it's genuinely covered, not just asserted to be.
- Grepped the whole tree for any other raw DID `==` comparison this PR might have missed — there's exactly one (`DidMapping::findIndex`), and this PR is the only place that pattern existed.

## Notes, not blockers
- `setMapping()` with a re-rendered DID updates the extension in place but does **not** overwrite the stored `.did` string to the new rendering (row keeps its original spelling, e.g. stays `+15551234567` even if later re-set via `"(555) 123-4567"`). Not discussed in the PR body but I think it's the right call — it avoids an accidental precision loss (dropping a `+`-qualified DID down to a bare national one) if an operator re-types carelessly. Worth being deliberate that this is the chosen behavior if anyone asks later.
- The accepted false-positive is inherently a table-wide risk, not just a pairwise one — two *different*, legitimately-configured DIDs (say a UK and a US number) that only collide via the relaxation would resolve via whichever was inserted first (`findIndex`'s linear scan). This is the same tradeoff already explicitly accepted and documented for the pairwise case, just extended to N>1 rows in an 8-entry table — not a new risk, just flagging that it's the same one at a different scope.

Genuinely strong work — the boundary documentation (stating the one accepted false positive plainly, with the reasoning for why the alternative is worse) and the mutation-testing discipline are exactly what this kind of routing-correctness code needs.

Issue #218, opened 2026-09-14
Music on hold does not play on an anchored/trunk call — hold is refused outright (confirmed on hardware)


GlomarGadaffi:
## Observed on real hardware

Live trunk call, `.244`, 2026-09-14: answered an inbound call, talked with no problem,
pressed hold, then retrieved — the caller heard **no music on hold** during the hold.
Retrieval itself worked cleanly (no dropped call, no error).

## Root cause: hold on an anchored/trunk call is refused outright, not silently unMoH'd

`onReinvite()` (`RequestsHandler.cpp:6666-6682`) and `onUpdate()`
(`RequestsHandler.cpp:6771-6780`) both refuse a hold re-INVITE/UPDATE with
`488 Not Acceptable Here` whenever the dialog's peer is `777` (echo), `888`
(conference), **or `kAnchorCallExt`  — the anchor extension every trunk/vendor-API call
rides** (`RequestsHandler.cpp:6672`, `:6772`):

```cpp
if (destNum == "777" || destNum == ConferenceRoom::EXT || destNum == kAnchorCallExt || !src || !dest)
{
	...
	response->setHeader("SIP/2.0 488 Not Acceptable Here");
	...
	return;
}
```

So the phone's hold re-INVITE never reaches anything that could act on it. The PBX's
own `Session::State` never leaves `Connected` — from the board's point of view the call
was **never actually held** — which is exactly why "hold, then retrieve" worked without
incident: nothing was released to retrieve. It also means the far end (in this case, an
outside caller reached via the trunk) may have kept hearing whatever the near end's
handset didn't locally mute, since the PBX-level media path never changed.

## This is a known gap, but a mislabeled one, and hardware now confirms it

Issue #194 (the SD/MoH/recording/voicemail epic) already documents this as the "Stage
2b" blocker:

> hold is currently refused with `488 Not Acceptable Here` whenever the peer is `777`,
> `888`, the anchor extension, or either leg is missing... Holding an
> anchored/trunk call does not work today at all.

Two things worth separating out, which the epic's single bucket obscures:

1. **`777`/`888` refusing hold is close to correct-by-design** — those are virtual
   legs with no real second party's media to hand back on resume, so there's nothing
   coherent to hold.
2. **The anchor extension (`555`, every trunk call) is a different case entirely, and
   the one worth fixing first.** For an anchored call the board already owns *both*
   RTP legs via `MediaBridge` — it is already in the media path, right now, for exactly
[...]

PR #238, opened 2026-09-15
SDP: RFC 8866 model + RFC 3264 offer/answer replace the six-field scanner (#196)

### LINKED ISSUE #196: SDP: replace the six-field line scanner with an RFC 8866 model + RFC 3264 offer/answer
Blocks three of the five stages in #194. Filing separately because it is a dependency those stages sit on top of, not part of any one of them.

## The problem

pocket-dial has no SDP model. It has a **six-field line scanner**, and SDP knowledge is split across two classes with the generic SIP message parsing SDP attributes — a layering inversion.

| Where | What it knows |
| :-- | :-- |
| `SipSdpMessage` | six **session-level** fields as cached offset spans — `v=` `o=` `s=` `c=` `t=` `m=` — and *"ignores `a=` attributes entirely"* (`SipSdpMessage.hpp:64-65`) |
| `SipMessage` (the base class) | `getSdpDirection()` (`SipMessage.hpp:187`), `filterAudioCodecs()`, `enforceG711()`, `checkSdp()` |

The only dedicated mutator is `setMedia()` (replaces the `m=` line). `c=` is read-only (`getConnectionInformation()`). Everything else goes through wholesale `setBody()` replacement, which is used in production (`ParkOrbit.cpp:44`, `:116`; `CallPickup.cpp:97`, `:110`).

Note also that the code cites **RFC 4566** throughout (10 mentions) — the superseded SDP spec. **RFC 8866** obsoletes it.

## What is actually wrong, concretely

1. **Exactly one `m=` section is assumed.** `SdpSpans::media` is a single `FieldSpan` (`SipSdpMessage.hpp:79`). A handset offering audio + video, or audio + a `telephone-event`-bearing second stream, is parsed as if the first `m=` were the only one.
2. **No media-level `c=`.** RFC 8866 §5.7 permits `c=` at session level *and* inside each media section, with the media-level value overriding. A phone that emits `c=` only per-media defeats `getConnectionInformation()` outright — and that is a legal, common offer.
3. **No attribute model.** `rtpmap`/`fmtp`/`ptime` are handled by line-passthrough rebuild: `filterAudioCodecs` passes every non-`a=rtpmap`/`a=fmtp` line through verbatim, so `c=` survives it unchanged by accident rather than by design.
4. **No construction or mutation API.** Anchoring media requires rewriting `c=` and the `m=` port **per leg**. Today that means hand-assembling a whole body string.
5. **Offer/answer rules are not enforced.** The unicast 200 OK is relayed with **no codec admission check at all** (`RequestsHandler.cpp:3346-3349`), while the caller's INVITE *is* gated with a 488 (`:1177`). A callee can therefore answer with a codec set the PBX would have rejected in an offer — and the PBX passes it on.

## What it costs today

- **Hold on anchored calls is refused outright.** `onReinvite` returns `488 Not Acceptable Here` when the peer is `777`, `888`, the anchor extension, or either leg is missing (`RequestsHandler.cpp:5416-5432`, duplicated in `onUpdate` at `:5521-5530`). That is exactly the set of calls where the board *is* in the media path.
- **G.722 is admitted but unusable on any board-terminated leg.** The relay gate allows PCMU/PCMA/G.722 (`:1177`), but every media object is PCMU-only: `RtpReceiver` drops non-PT-0 (`:359-361`), `RtpSender` always emits PCMU (`:344`, `:528`), `buildMediaSdp` hardcodes `RTP/AVP 0` (`:1536-1537`).
- **A stale hold SDP can be spliced into a later transfer.** The relayed answer is stored with `setRemoteSdp()` (`:3200-3201`) and reused to build attended-transfer offers (`:3671-3674`); the code flags the hazard itself at `:3666-3668`.

## Scope

**RFC 8866** — a real parse/serialize model: session and media levels, multiple `m=` sections, media-level `c=` override, an attribute model covering at least direction / `rtpmap` / `fmtp` / `ptime`, and construction + mutation rather than string assembly.

**RFC 3264** — the offer/answer half, which is what actually governs anchoring: what constitutes a valid answer, re-offer rules, hold semantics (§8.4), and rejecting a stream with port 0.

These are two different specs and both are needed. 8866 defines syntax and semantics; 3264 defines the conversation.

## Three constraints any implementation must respect

The current design is defensive, not naive. A rewrite that ignores these makes things worse.

1. **Parsing must stay flat and non-recursive, bounded by input length rather than by structure.** `SipSdpMessage.hpp:59-69` documents why `a=` is ignored today: the UNISOC T612 VoLTE RCE (CWE-674) was uncontrolled recursion in an `a=acap` decoder, where a body of repeated `acap:1 acap:1 …` recursed per token until the modem stack overflowed into a neighbouring task. Introducing attribute parsing reintroduces that attack surface unless it is explicitly flat. Keep `SdpLimits::kMaxLines` and `checkSdp()` in front of it.
2. **Cache offsets, never `string_view`s.** The message pool recycles slots with `*msg = source`, so a cached view would be copied verbatim and left dangling into the source message's body. The existing offset/generation scheme (`_spans` / `_spansGen`) exists for exactly this reason and must survive.
3. **Fixed capacity, no heap growth.** `PoolConfig.hpp` discipline. A `std::vector<Attribute>` per media section is not acceptable on this part — use fixed-capacity arrays with documented caps and defined overflow behaviour.

## What this unblocks in #194

| Stage | Why it needs this |
| :-- | :-- |
| **2b** — MoH on phone-initiated hold | Must set direction attributes and answer as a UAS mid-dialog |
| **3** — Voicemail | Must answer with constructed SDP and survive re-INVITE/UPDATE |
| **4** — Recording / B2BUA | Must rewrite `c=` and `m=` port per leg, and keep hold/resume/transfer coherent |

**Stage 1 (SD-backed CDR archive) does not depend on this** and can proceed in parallel.

## Suggested sequencing

Model and parser first, with the existing six accessors reimplemented on top of it so nothing else changes, and the existing hardening tests (`SipSdpMessage_hardening_test.cpp`) kept green. Then offer/answer rules. Then the anchoring work in #194 Stage 4.

Refs #194

GlomarGadaffi:
## Summary

Replaces the six-field SDP line scanner with an RFC 8866 model and an RFC 3264 offer/answer engine, and wires them into the four places the engine's behaviour was wrong. Model and parser first, the six existing accessors reimplemented on top, then offer/answer, then the handler touch points, as #196 suggested.

**New**
- `src/SIP/Sdp.hpp/.cpp` (`namespace sdp`): fixed-capacity, offset-based model of a whole body: session fields, every `m=` section (up to 3), per-section attribute scoping (RFC 8866 §5.7/§5.13: media-level `c=`/direction override the session's), `rtpmap`/`fmtp`/`ptime` decoded, everything else kept as opaque spans. Queries: `effectiveDirection`, `effectiveConnection`, `connectionAddress` (with `0.0.0.0` detection), `isHold` (RFC 3264 §8.4 plus the RFC 2543 zero-address form), `telephoneEventPt`. A writer that replaces the three hand-assembled bodies with byte-identical output (pinned by tests).
- `src/SIP/SdpOfferAnswer.hpp/.cpp`: `buildAnswer` (one `m=` per offered `m=` in offer order; non-audio or no-common-codec streams answered with port 0 instead of silently dropped; formats = offer ∩ local caps in the offer's order; the offered telephone-event PT echoed with the *offered* fmtp; direction = RFC complement limited by what the board wants), `validateAnswer` (§6/§6.1: m= count, order/types, proto, formats ⊆ offer), `holdRequested`, `answerDirection`.

**Wired in**
1. `SipSdpMessage`'s six accessors now derive from the model: `getMedia()`/`getRtpPort()` report the **first audio** section (a video-first offer no longer reads the video line as "the" media), `getConnectionInformation()` is the c= that **effectively applies** to it (media-level override). `parseCallerRtp()` inherits both fixes.
2. `SipMessage::getSdpDirection()` judges the first audio stream with scoping, and reports `c=0.0.0.0` as `Inactive`, so `onReinvite`/`onUpdate` now treat legacy hold as hold. `None` still means "no attribute anywhere".
3. The callee's 200 OK was relayed unread while the caller's INVITE was 488-gated (#196 item 5). Both onOk relay branches now run `validateAnswer` against the stored INVITE **before any state moves**; an answer that lists an un-offered payload type or drops an offered stream is dropped, and the callee's retransmit or corrected answer is judged afresh.
4. 440/888/555 answer via `answerSdpFor()` → `buildAnswer` when an offer is present: an audio+video offer now gets `m=video 0 …` in the right position instead of our audio matched against the phone's video port, and DTMF is echoed with the phone's own fmtp range on 440 too.

**The three constraints from the issue, and how they are met**
- Flat and bounded: `sdp::parse` is one pass, no recursion, no token re-dispatch, capped by `SdpLimits::kMaxLines` and its own fixed capacities; overflow is counted in `dropped*`, never decoded. `tests/tools/check_parser_callgraph.py` (now compiling the two new units, with `sdp::parse/buildAnswer/validateAnswer` as roots): no cycles, no dynamic frames.
- Offsets, never views: `Span{pos,len}`; `sdp::view()` resolves against the body and returns empty on mismatch.
- No heap growth: `Session` is a trivially-copyable aggregate of fixed arrays. It is ~1.7 KB, which is too big for the **8 KB `sip_server_task` stack** and too big to cache per pooled message, so two engine-owned static scratch Sessions are parsed on demand under the existing `_mutex` contract. The guard caught `out = Session{}` materialising a 2.2 KB temporary in `parse()`'s frame; it now clears in place (frame 144 bytes). `buildAnswer`'s frame is 1.3 KB (its `SessionSpec` and echo buffers), on par with existing handler frames.

**Deliberate scope limits**
- The board's media objects are PCMU-only (`RtpSender`/`RtpReceiver`), so `LocalCaps::pcmuOnly()` is what 440/888/555 answer with. A PCMA-only offer to a service extension therefore still falls back to the old PCMU answer rather than newly 488ing those handsets; the engine already returns the honest port-0 answer (`SdpOfferAnswer.PcmaOnlyPhoneAgainstPcmuOnlyBoardIsRejectedNotLiedTo`), so switching to a 488 is a one-line policy change once the media path speaks PCMA/G.722.
- The virtual-leg 488-on-hold branch in `onReinvite`/`onUpdate` is untouched (#218 is in flight on it).
- Cross-reference to #239 (#218 hold answer): `answerSdpFor()` complements the offered direction, so the sendrecv-vs-recvonly gap flagged on the 555/anchor hold answer in #239 resolves as a side effect once both are in; whichever lands second should re-check that path.
- Server-originated *offers* (register beep, park, MoH preview) still use `buildMediaSdp`/`makeInactiveHoldSdp`; the writer reproduces them byte for byte, so migrating them is optional.
- Anchoring's per-leg `c=`/port rewrite (#194 stage 4) is the next consumer of the writer, not part of this PR.

**Behaviour change pinned in an existing test:** `SipSdpMessageCache.LineEndingAndDuplicateFieldBehaviorUnchanged` asserted last-one-wins for two `m=audio` lines. Two `m=` lines are two streams (RFC 8866 §5.14); the accessors now report the first audio section, and the test says so.

## Tests
- `tests/Sdp_test.cpp` (14): multi-m=, media-level c=, direction scoping/default, `0.0.0.0` hold, bare LF, counted overflow, acap flood bounded, reparse reset, span/body mismatch, port-0 and `/2` ports, writer byte-identical to `buildMediaSdp` and `makeInactiveHoldSdp`, writer→parser round trip.
- `tests/SdpOfferAnswer_test.cpp` (7): PCMA-first answered in offer order, PCMA-only vs PCMU-only board → port 0, video rejected in place and self-validates, direction complement table, hold answered recvonly, `validateAnswer` accept/reject matrix, `holdRequested` on direction/session-level/zero-address/video-first bodies.
- `tests/SdpAdmission_test.cpp` +3 through the real `handle()` path: 440 answers audio+video with the video stream rejected in position and Content-Length correct; a relayed 200 OK naming G.729 (never offered) is dropped and a corrected retransmit relays; an answer one `m=` short is dropped and a port-0 rejection relays.
- All pre-existing SDP suites (`SdpAdmission`, `SdpAdmissionGate`, `SdpNegotiate`, `SipSdpHardening`, `SipSdpMessageCache`, `DecodePathsAllocateNothing`) green unchanged apart from the one assertion above.

**820/820 host tests green under WSL** (re-run after the review fixes: dead clock ternary removed, `validateAnswer` compares modelled and dropped m= counts separately so sections always align index-for-index). Call-graph guard clean. Firmware build not run in this session; nothing here is behind an `ESP_PLATFORM` arm. Hardware: untouched.

Closes #196


GlomarGadaffi:
Vetted independently (same GitHub account as the PR author, so this is a comment not a formal approval — verdict: **ship it**).

## What I checked
- Read `Sdp.hpp`/`.cpp` and `SdpOfferAnswer.hpp`/`.cpp` in full, plus all four `RequestsHandler` wiring touch points and the `SipMessage::getSdpDirection()`/`SipSdpMessage::ensureParsed()` accessor rewrites.
- Built the PR branch from scratch in an isolated worktree (not trusting the PR's own numbers): **820/820**, matches exactly.
- Independently ran `tests/tools/check_parser_callgraph.py --report` myself and confirmed the exact frame-size claims: `sdp::parse()` is **144 bytes**, `sdp::buildAnswer()`'s own frame is **1344 bytes** (~1.3 KB) — not just asserted, verified. Also checked the deepest static chain reported at 6416/8192 bytes (`RequestsHandler::handle()` → `BlfSubscriptions::buildDialogNotify`) — that's pre-existing, unrelated to this PR (BLF notify XML building, nothing to do with SDP), so this PR doesn't touch the tightest existing margin.
- Ran the 46 new/touched SDP tests in isolation (`Sdp.*`, `SdpOfferAnswer.*`, `SdpAdmission.*`) — all pass, including the real `handle()`-path integration tests that show the actual log lines ("answer lists a payload type the offer did not", "answer m= count differs from offer").
- Mutation-tested the most security-relevant guard myself rather than trusting the description: neutralized `validateAnswer()`'s `FormatNotOffered` check (`if (false && !o.hasFormat(...))`), rebuilt, reran. Caught by **two** tests — the unit test AND `SdpOfferAnswerGate.RelayedAnswerListingAnUnofferedCodecIsDroppedNotRelayed`, which demonstrates the actual consequence: with the guard gone, the bad 200 OK *is* relayed to the caller and the call *does* connect (`session.value()->getState()` moves off `before`). Reverted. This is real protection with a demonstrated failure mode, not just a claim.
- Traced `getInviteMessage()`'s safety against pool reuse: it's `Session::_inviteMessage`, a `shared_ptr` held directly on the `Session` (pre-existing pattern, already used at half a dozen other call sites), not a pool re-lookup by call-ID — so the stored offer can never be clobbered by an unrelated call reusing that pooled `SipMessage` slot while the session is still live. This matters because `validateAnswer()` at the `onOk()` touch points depends on comparing against the *correct* original offer.
- Hand-traced `getSdpDirection()`'s new logic against the old flat-scan version: correctly does NOT just delegate to `sdp::effectiveDirection()` (which RFC-defaults an unstated direction to `SendRecv`) because this caller needs to keep distinguishing "no attribute anywhere" (`None`) from "explicitly `sendrecv`" for backwards compatibility — that's a deliberate, correct divergence, not redundant duplication.

## Notes, not blockers
- `SdpOfferAnswer.cpp:106`: `f.clock = (pt == 9) ? 8000 : 8000;` — both branches are the same value, so the ternary is a no-op today (it's documenting the RFC 3551 G.722-clock-is-nominally-8000-despite-16kHz-sampling quirk, not implementing a real branch). Not a bug — the output is correct for every codec this handles — but it reads like a branch was meant to differ and didn't. Worth either a plain `= 8000;` with the quirk explained in a comment, or leaving as-is if the ternary shape is intentionally future-proofing for a codec that *would* need a different clock.
- `validateAnswer()`'s per-section loop only runs over `min(offer.mediaCount, answer.mediaCount)`, gated by an earlier check that only compares *totals* (`mediaCount + droppedMedia`). In the narrow case where an offer and a relayed answer both exceed `kMaxMedia` (3) but distribute their "modeled vs. dropped" sections differently while the totals happen to match, one of the offer's actually-modeled sections could go unchecked against the answer. Requires a 4+ `m=` line SDP body (already an exotic shape for this PBX) and only weakens validation of a peer-to-peer relayed call's media the board doesn't decode anyway — low severity, not worth blocking on, but worth a one-line comment or a follow-up test if `kMaxMedia` is ever raised.

Excellent work — the offset-based flat model is exactly the right structural answer to the CWE-674 lesson this codebase already carries, and the discipline of pinning the writer byte-identical to the old hand-assembled builders means this is a genuinely safe drop-in, not a rewrite-and-hope.

GlomarGadaffi:
Superseded by #255, which redid #196 Phase 2 from the original stricter spec (re2c-generated lexer, zero std::string/std::vector, borrowed static scratch, ASAN dangling-pointer coverage) and merged as d410094. Nothing here is being salvaged into that PR.

Issue #166, opened 2026-09-13
911: not dialable specially today, no correct failure code with no PSTN route, no Kari's Law notification hook


GlomarGadaffi:
## Current behavior (verified, not assumed)

There is no special handling of \`911\` (or \`933\`, the standard US E911 test number)
anywhere in the codebase — \`grep -rn '\"911\"' src/\` returns nothing. Dialing \`911\`
today goes through ordinary dial-plan matching like any other number:

- If no extension is registered as \`911\` and no dial-plan rule matches it, the call
  gets a plain **\`404 Not Found\`** (\`CallForker.cpp:397\`, \`routeDialPlan()\`'s
  no-match fallback, confirmed by reading the code) — the same response as dialing
  a typo'd internal extension. Nothing distinguishes \"this number doesn't exist\"
  from \"this is 911 and there's no way to reach emergency services from here.\"
- If a device *does* happen to register as \`911\` — nothing stops that today, see
  the companion reserved-extension-guard issue — the call would simply ring that
  phone. No safeguard, no warning, nothing PSTN-related happens at all.
- There is no PSTN routing at all reachable from ordinary number-dialing yet — see
  the companion dial-plan/E.164 issue; \`555\` is the only way to reach
  \`TelephonyAnchorClient\` today, and it carries no destination number.

## What's needed (tracking placeholder — needs a design pass, including telecom/
regulatory input, before implementation)

1. **\`911\`/\`933\` always dialable, bypassing whatever dial-plan lock/restriction
   state might otherwise apply** (e.g. a \`secure\`-mode registrar or a restrictive
   dial-plan pattern must never block 911 specifically).
2. **Preferentially routed to whichever PSTN trunk is currently active**
   (TelephonyAnchorClient today; a SIP ITSP trunk once #164 lands) — this is a
   hard dependency on the dial-plan trunk-routing action from the companion issue
   existing at all.
3. **A correct SIP failure response when no PSTN trunk is configured/reachable.**
   Today's blanket \`404\` is definitely the wrong signal for 911 specifically — it
   reads as \"no such destination\" to the calling UA rather than \"this device
   cannot currently reach emergency services.\" What the *correct* replacement code
   is (\`503 Service Unavailable\` is the common convention for \"can't complete this
   call right now,\" but this deserves a real citation/RFC-6881-territory check or
   input from someone with telecom-regulatory background, not a guess baked into
   an issue as settled) is an open question for whoever picks this up.
[...]

Issue #163, opened 2026-09-13
REGISTER admits reserved/emergency/PSTN-shaped identities — no reserved-extension or E.164 guard on the registrar


GlomarGadaffi:
## Problem

\`RequestsHandler::onRegister()\` (src/SIP/RequestsHandler.cpp:819) validates the
From-number against exactly one check: \`isValidAor()\` (RequestsHandler.cpp:5200),
which is a **charset-only** filter (alnum + \`.-_+*#\`). It applies in every registrar
mode (Open, Secure, Learn) — there is no identity-semantics check anywhere in the
REGISTER path.

Confirmed by reading the code, not assumed:

- Nothing stops a device from registering as one of the reserved virtual extensions
  already load-bearing elsewhere in the engine: \`777\`/\`999\`/\`888\`/\`555\`/\`440\`
  (echo test, DTMF admin/CLASS gate, meet-me conference, the anchor media bridge
  \`kAnchorCallExt\`, and busy/error tone respectively — see RequestsHandler.cpp:958-1314
  and PbxFeatureConfig.cpp:98/162/285 for the existing, scattered copies of this
  literal set). A rogue or misconfigured phone claiming \`555\` would collide directly
  with the anchor-call virtual extension.
- Nothing stops a device from registering as \`911\` (or \`933\`, the standard US E911
  test number) — there is **zero** special handling of either anywhere in the
  codebase (\`grep -rn '\"911\"' src/\` returns nothing). A phone that registers as
  \`911\` would simply receive calls dialed to that number instead of anything
  emergency-relevant happening.
- Nothing stops a device from registering as an E.164-shaped string (e.g.
  \`+15551234567\`) — the charset explicitly allows \`+\`, and there's no length/shape
  check distinguishing \"internal extension\" from \"direct PSTN number.\" \`+\`-prefixed
  strings are unambiguously E.164; unprefixed all-digit long strings are a policy
  call (see below).

## Why this matters now

This surfaced while designing 911 handling (see the companion issue for full 911
scope) — an attacker or a misconfigured device registering as \`911\` before a real
emergency-routing rule exists would silently intercept anything dialing \`911\`, with
no indication anything is wrong.

## Fix (scoped, mechanical)

Add an identity guard in \`onRegister()\`, evaluated **before** the mode branch
(~line 846) so it applies in Open/Secure/Learn alike:

[...]

PR #241, opened 2026-09-15
911/933 dialing: resolved before the dial plan, and 503 instead of 404 (#166 part 1, fixes #240)

### LINKED ISSUE #240: A "9*" outside-line dial rule rewrites a dialed 911 into 11
## Summary

A dial-plan trunk rule written the ordinary way — pattern `9*`, `stripDigits=1`, i.e. "dial 9 for an outside line" — **matches a dialed `911` and rewrites it to `11`** before handing it to the trunk. The call is placed to `11`. Nothing errors, nothing logs a warning, and the operator has no indication their outside-line rule swallowed an emergency call.

Found while implementing #166. Present in `main` and in the shipped `v1.5.0-beta.2` firmware.

## Why it happens

Two correct-in-isolation behaviours compose badly.

`dialPatternMatches()` (`src/SIP/DialPlan.hpp`) treats a trailing `*` as "absorb the rest":

```cpp
const bool prefixMode = (pattern.back() == '*');
const size_t litLen = prefixMode ? pattern.size() - 1 : pattern.size();
if (!prefixMode && dialed.size() != litLen) return false;
if (dialed.size() < litLen) return false;
```

So pattern `9*` has `litLen == 1`, and `911` matches on its first character alone.

`applyTrunkTransform()` (same file) then strips unconditionally:

```cpp
out = prepend + dialed.substr(static_cast<size_t>(stripDigits));
```

`"911"` minus one leading digit is `"11"`. `DialRuleTable::match()` is first-match-wins over an ordered vector, so a later, more specific `911` rule never runs.

## Reproduce

Host suite, no hardware:

```cpp
handler.setDialRule("9*", "trunk", "", 1);
handler.handle(invite(/*from=*/"101", /*to=*/"911", ...));
// before the fix: lastMakeCallDestination() == "11"
```

## Scope of exposure

- Only bites a deployment that has configured a wildcard trunk rule whose pattern matches `911`. The rule on the `.244` bench is `9XXXXXXXXXX` (11 characters, exact-length), which does **not** match a 3-digit `911` — which is why this was never seen there.
- Any deployment using the shorter, more natural `9*` form is affected.
- `933` (the E911 test number) is affected identically.

## Fix

#166's emergency intercept, which resolves `911`/`933` **before** the dial plan and before every other operator-configurable destination lookup (ring groups included), and always hands the trunk the bare `911` regardless of what was dialed. A dial-plan rule can no longer see an emergency number at all.

Two related hardening items land with it:

- `PbxFeatureConfig.cpp` carried **three different** hand-copied reserved-extension reject lists (ring groups and forwards omitted `440`; the dial-plan validator omitted `888`; **none** covered `911`/`933`). They now all call `pbx::isReservedExtension()`, which was previously reachable only from the REGISTER identity guard.
- A ring group named `911` was accepted and persisted. `findRingGroup()` runs before the dial plan in `onInvite()`, so that group would have shadowed emergency dialing outright.

## Note

This is a defect in how an emergency call is routed, so it is worth reading alongside `docs/FEATURE_ROADMAP.md`'s standing disclaimer: this firmware should not be anyone's only means of calling for help.

GlomarGadaffi:
Fixes #240. Part 1 of #166 — the Kari's Law notification hook is part 2.

## There was a live bug here, and it is the reason this is worth reading carefully

A dial-plan trunk rule written the ordinary way — pattern `9*`, `stripDigits=1`, i.e. "dial 9 for an outside line" — **matches a dialed `911` and rewrites it to `11`.**

`dialPatternMatches()` treats a trailing `*` as "absorb the rest", so `9*` has `litLen == 1` and matches `911` on its first character alone. `applyTrunkTransform()` then strips one leading digit. `DialRuleTable::match()` is first-match-wins over an ordered vector, so a later, more specific `911` rule never runs. The call is placed to `11`. Nothing errors and nothing logs a warning.

The `.244` bench rule is `9XXXXXXXXXX` — 11 characters, exact-length — which cannot match a 3-digit `911`. That is why this was never seen on the bench. Any deployment using the shorter, natural form is affected, including in the shipped `v1.5.0-beta.2`. Filed separately as #240.

## Where the check goes, and why that placement *is* the feature

The fix is deliberately **not** a dial-plan rule. The dial plan is operator-editable and first-match-wins, so any rule can be shadowed by one ordered above it.

The emergency check is now the **first destination check in `onInvite()`** — ahead of the reserved virtual extensions (777/999/440/888/555/70x), ahead of `findRingGroup()`, and ahead of `routeDialPlan()`. No configuration an operator can type is capable of shadowing, reordering or rewriting it.

It also sits deliberately **above** the secure-mode INVITE challenge while staying **below** the registration and codec gates. That line is drawn on a principle rather than by feel:

| gate | kind | applies to 911? |
|---|---|---|
| caller is registered (403) | capability | **yes** — an unidentifiable caller has no working call to place |
| codec this PBX can relay (488) | capability | **yes** — otherwise 911 connects with dead audio |
| secure-mode INVITE challenge | policy | no |
| ring groups | policy | no |
| dial plan | policy | no |
| CFU / DND | policy | no |

**Capability gates stay, policy gates do not.** That satisfies #166 item 1 ("bypassing whatever dial-plan lock/restriction state might otherwise apply") without turning the PBX into an unauthenticated 911 originator for anything that can reach UDP/5060.

## The trunk is always handed the bare number

One leading trunk-access digit is absorbed, because a user on a "dial 9 first" system will dial 9-1-1 out of habit, and FCC 19-76 permits supporting that alongside direct dial (it is *direct* dial that is mandatory). So:

- `911` → trunk gets `911`
- `9911` → trunk gets `911` — never `9911`, never `11`
- `933` → trunk gets `933`, tagged TEST in every operator-facing string

The classifier is an exact match against a closed set with one optional prefix digit. No prefix matching, no "starts with 911", no length guessing: a false positive hijacks an ordinary call onto the emergency path, and a false negative drops a 911 call into normal routing.

## 503, and specifically not 404

With no trunk connected the answer is now **503**, with `Warning: 399` and deliberately **no `Retry-After`**.

#166 item 3 asked for "a real citation/RFC-6881-territory check ... not a guess baked into an issue as settled". So:

- **RFC 4497 §8.3.1 (BCP 117)** covers this exact condition — a SIP INVITE inbound with no outbound channel available: *"If no suitable channel is available, the gateway should use response code 503 (Service Unavailable)."*
- **RFC 3398 §7.2.4.1**'s Q.850 mapping agrees: every trunk-outage cause (34 no circuit, 38 network out of order, 41 temporary failure, 42 congestion, 47 resource unavailable) → 503, while 404 is reserved for numbering-plan causes (1/2/3).
- **404 is not merely a worse choice, it is a false statement.** RFC 3261 §21.4.5 defines it as *"the server has definitive information that the user does not exist"* — but the PBX just recognised 911. It knows exactly what the number is; only the trunk is missing.
- **No `Retry-After`**, per RFC 3261 §21.5.4: a client *"SHOULD NOT forward any other requests to that server for the duration specified"*. Making a handset back off the PBX after a failed 911 attempt is the last thing anyone wants.
- **RFC 6881 specifies no code at all**, because §8 SP-28 makes a default mapping a MUST and so never contemplates a compliant proxy having no route. 503 is the least-wrong answer to a state the BCP forbids.
- There is no emergency-specific SIP warn-code (IANA registry checked); 399 is the only fit.

This was researched with parallel agents and then **adversarially verified — two independent verifiers re-fetched the RFCs from rfc-editor.org and checked every cited section; 0 of 2 refuted.**

`originateAnchorCall()` is called with `respondIfDisconnected=false` so it stays silent and this path owns the response. The handset never receives two final responses to one INVITE, and a test asserts exactly that.

## Reserved-extension validation was three lists that had drifted apart

`PbxFeatureConfig.cpp` carried three separately hand-copied reject lists:

| validator | had | missing |
|---|---|---|
| `setForwardLocked` | 777/999/888/555 | 440, **911**, **933** |
| `setRingGroup` | 777/999/888/555 | 440, **911**, **933** |
| dial-plan pattern | 777/999/440/555 | 888, **911**, **933** |

`pbx::isReservedExtension()` has had the complete list all along and was reachable only from the REGISTER identity guard. All three now call it.

Concretely: a ring group named `911` was accepted and persisted. `findRingGroup()` runs before the dial plan in `onInvite()`, so such a group would have shadowed emergency dialing outright had the intercept not moved ahead of every lookup. This is defence in depth, not the fix.

Also adds `SipMessageTypes::SERVICE_UNAVAILABLE`. Every other status had a constant and 503 did not, so it was a bare literal at six-plus call sites — part of why the no-trunk path reached for `NOT_FOUND` in the first place.

## Testing

**15 new tests.** Roughly half assert what must *not* happen, because both failure directions here are severe.

Every guard was mutation-tested — including **relocating the intercept below the dial plan**, to prove the ordering test actually tests ordering rather than merely passing:

| mutation | caught by |
|---|---|
| intercept relocated below the dial plan | `AnOutsideLineRuleCannotRewriteNineOneOneIntoEleven` |
| classifier never fires for 911 | `DialingNineOneOneReachesTheTrunkWithTheBareNumber` |
| trunk-access digit no longer absorbed | `DialingNineNineOneOneStillReachesTheTrunkAsNineOneOne` |
| dialed string handed to trunk instead of bare number | `DialingNineNineOneOneStillReachesTheTrunkAsNineOneOne` |
| no-trunk answer reverted to 404 | `WithNoTrunkConnectedTheAnswerIsFiveOhThreeAndNeverFourOhFour` |
| `Retry-After` added to the 911 failure | `TheFiveOhThreeCarriesAWarningAndNoRetryAfter` |
| `originateAnchorCall` allowed to answer too | `ANineOneOneDialIsNeverAnsweredTwice` |
| ring-group validator back to its old list | `ARingGroupCannotBeNamedForAnEmergencyNumber` |

The suite was rebuilt after restoring each mutation, so none of this is a stale binary reporting green.

`RegisterIdentityGuard_test.cpp`'s 911 test asserted the old 404 and is updated. Its real assertion — a refused REGISTER binds nothing, so the squatter never receives the call — is unchanged and now stronger; but reaching the unclaimed-number 404 tail would mean the intercept was bypassed, so it now asserts the *absence* of that 404.

Full host suite green under WSL.

## What this does NOT do

- **The Kari's Law on-site notification hook** (#166 item 4). Part 2, so #166 stays open.
- **The same 404 → 503 correction for *ordinary* trunk calls** with no anchor connected (`originateAnchorCall`'s `respondIfDisconnected=true` branch, and `CallForker`'s shared no-match tail). The same RFC 4497 citation applies, but it changes non-emergency behaviour, and `CallForker`'s 404 tail is shared with genuinely-absent ring-group and page-zone targets where 404 *is* correct. That needs its own change and its own review rather than riding along here.
- **A way to customise 911 routing.** 911 now always goes to the active anchor. An operator who wanted a specific trunk target for it has no knob; that is the safe default, not a considered feature.
- **Anything verified against a real carrier.** The host suite proves the engine hands the trunk `911`; it cannot prove the trunk delivers it, and it says nothing about PSAP routing or dispatchable location — both of which live with the trunk provider. `docs/FEATURE_ROADMAP.md`'s disclaimer stands: do not deploy this as anyone's only means of calling for help.

## Hardware

Untouched. Host suite only; no flash.


GlomarGadaffi:
Vetted independently alongside #242 (same GitHub account as the author, so a comment not a formal approval — verdict: **ship it**).

## What I checked
- Read `EmergencyCall.hpp`'s classification logic and the `onInvite()` wiring in full. Confirmed by direct code read (not just the PR text) that the intercept sits exactly where claimed: after the registration (403) and codec (488) capability gates, before the secure-mode challenge and everything dial-plan/policy-related.
- Built #242's branch (which includes this one, stacked) from scratch in an isolated worktree: **833/833**, matches Opus's number exactly.
- Hand-traced `classifyEmergencyDial()`: the trunk-prefix stripping only fires when the exact 4-char remainder is "911"/"933", so it can't misfire on an ordinary 9-prefixed PSTN number (which would never be exactly 3 digits after the 9) or a real 4-digit extension happening to start with 9.

## One real finding: a doc comment this PR makes false, in a file it doesn't touch
`PbxConfig.hpp`'s existing (#163-era, pre-this-PR) comment on `isReservedExtension()` explicitly says the three separate lists in `PbxFeatureConfig.cpp` (`setForwardLocked`, `setRingGroup`, `setDialRule`) "deliberately does NOT" get retrofitted onto the shared helper, and "must go on" omitting 911/933 "for a future dial-plan rule routing 911 to a trunk."

This PR does exactly the opposite: all three now call `pbx::isReservedExtension()` directly (confirmed in the diff and by `ReservedExtensionRefusalIsNowOneSharedListNotThreeDrifted`, which explicitly asserts 911/933 are refused as ring-group names and dial-rule patterns). I think the new behavior is *correct* — a dial-plan rule for 911 is now not just unnecessary but actively dangerous, since it would be redundant with (and could shadow reasoning about) the hardcoded intercept that's the whole point of #166 — but the `PbxConfig.hpp` comment is now a live, false claim about the codebase's own design intent that will mislead the next person who reads it. Worth a one-line update to that comment (not in this PR's file list, so easy to miss) noting that #166 superseded that rationale.

Nice, real catch in the drift itself: `setForwardLocked`/`setRingGroup` were missing 440, `setDialRule` was missing 888, and none of the three covered 911/933 before this. A ring group literally named "911" was accepted and persisted pre-#166.

PR #242, opened 2026-09-15
Kari's Law: on-site notification when 911 is dialed (#166 part 2)

### LINKED ISSUE #166: 911: not dialable specially today, no correct failure code with no PSTN route, no Kari's Law notification hook

GlomarGadaffi:
Closes #166. Part 2 of two — part 1 (the emergency intercept and the 503) is #241, which this builds on.

## What Kari's Law actually requires

47 CFR 9.16(b), from the Kari's Law Act of 2017, has two halves. Part 1 (#241) did the first: 911 must be reachable from any station with no prefix or code. This is the second: a notification must be sent when 911 is dialed, to a central location at the facility or to someone able to act on it. 47 CFR 9.3's "MLTS Notification" definition fixes what it carries — the fact a 911 call was made, a valid callback number, and whatever location information the system conveys to the PSAP (the latter two excused where technically infeasible).

**The rule text is paraphrased throughout, never quoted.** The research behind this could not retrieve verbatim eCFR text (302 redirects and 403s from eCFR, fcc.gov and 911.gov), so the behaviour here is the conservative reading that four independent summaries agreed on. Nothing in the code or comments should be cited as the letter of the law.

## The ordering rule *is* the compliance rule

9.16(b)(2) requires the notification to be contemporaneous with the call and to **not delay it**. That has an exact code form, and it is the property most of the tests assert:

> the emergency call leg is enqueued **first**, then the notification runs, and no failure path in the notification — pool exhaustion, an unregistered target, an empty or over-long config — can prevent, delay or reorder the call.

Because the call leg and the notifications land in the same `_outbox` and leave on the same `drainOutbox()` pass, "contemporaneous" holds *literally* rather than approximately. Two tests assert position on the wire, not merely presence:

- `TheCallLegIsOnTheWireBeforeTheNotification` — 200 OK precedes the MESSAGE
- `TheFiveOhThreeIsOnTheWireBeforeTheNotificationToo` — same on the failure path

## It fires when the call *fails*

A 911 attempt that went nowhere is the single most important thing to put in front of a human, and it is the case an implementation is most likely to miss. `notifyEmergency()` runs on all three paths — routed, answered-503, and even the pool-exhausted path where no response could be built at all — and the text says which happened. The rule is about a call being *placed*, not about it succeeding.

## Channels

| channel | guaranteed? | why |
|---|---|---|
| syslog, severity **Alert**, app `pbx-911` | **yes** — unconditional, unconfigurable | `Syslog::send` never blocks, never throws, no-ops when unconfigured, so it is safe on the SIP thread under the engine mutex. An operator can leave the notify list empty but cannot turn off the *record*. |
| SIP MESSAGE per configured notify extension | best-effort | enqueued on the same outbox pass as the call |
| intercom beep per notify extension | best-effort | the "or hear it" half |

**Two honest limitations, documented in the code rather than hidden:**

1. **No handset here has ever been sent a SIP MESSAGE.** This notifier composes its own (it cannot call `sendMessageTo()` — that takes `RequestsHandler::_mutex` and self-drains, so calling it from `onInvite()` would deadlock on a non-recursive lock). The wire format is now tested; whether a given handset *renders* it to a human is not, and cannot be on host. That is why the notification does not rely on MESSAGE alone.
2. **The beep is `RegisterBeeper::sendBeep()`**, reused because it is the one server-originated "make this phone make a noise" path that is actually proven in production. It *skips when its bounded table is full*, which is right for a cosmetic registration beep and is not something anyone should rely on for 911. It is strictly additive; the syslog record and the MESSAGE are the notification. A dedicated alert-INVITE table that cannot be starved by registration churn is the follow-up.

The dialing extension is never beeped even if it is on the notify list — it is mid-call-setup and an intercom INVITE at that moment is the last thing it needs.

## Structure

`EmergencyNotifier` is a sibling machine in the `CallForker`/`RegisterBeeper` mould, reaching the engine only through `PbxEnv`. That was deliberate: `RequestsHandler.cpp` is the most contended file in the repo and #218 is in flight there, so this adds **one member and one call** to it rather than more surface. The notification text is a pure function (`pbx::formatE911Notification`) so it is host-tested without a handler, a socket or a clock.

Config lives in `PbxFeatureConfig` under its own NVS key `e911`, so a rewrite of any other table cannot quietly drop who gets told when somebody dials 911. `setE911Config()` is deliberately **total** — it drops an unusable field and keeps the rest rather than refusing the write — because a rejected config would leave the *previous* notify list silently in place while the operator believed they had changed it.

## Config surface

`GET`/`PUT /api/e911-config` (`notifyExts`, `callback`, `location`), matching the existing telephony/DID config model: read at admin, write at sysop. `PUT` echoes back the stored result so a caller sees what actually survived truncation and field-dropping.

- **callback**: nothing in this firmware supplies one — the anchor's `routeDn` is an origination DN and `DidMapping` is inbound-only — so it is operator-set. When empty the notification says `callback not configured` explicitly rather than omitting the field, because a blank field reads as "nothing to report" and here it means the opposite.
- **location**: a free-text site string. This is the cheapest honest step toward a dispatchable location. **It is not sent to the PSAP** and is not presented as if it were.

## Testing

22 new tests, **833 total**, full host suite green under WSL. Guards mutation-tested, including inverting the ordering (moving the notification above `originateAnchorCall()`) to prove the ordering tests actually test ordering.

**One mutation survived the first pass and that was worth chasing.** Removing the notify-list bound *inside* `EmergencyNotifier::notify()` changed nothing observable, because `setE911Config()` already truncates before the notifier ever sees the list. So the notifier's own bound is defence-in-depth against a config this build did not write, and unreachable through the public API by construction. Rather than delete a guard that protects the emergency path from an unbounded burst of pooled messages, or leave it permanently untested, the notifier is now driven directly through a counting `PbxEnv` and the mutation is caught.

### The MESSAGE wire format is now actually exercised

`sendMessageTo()` having zero callers does not just mean one dead function: it means **nothing in this firmware has ever put a SIP MESSAGE on the wire at all**. This notifier is the first thing that will, and its builder was written by reading the never-called one, which is how a format bug gets inherited rather than caught.

Four tests close that: request line and method, the Via `z9hG4bK` magic cookie (RFC 3261 §8.1.1.7), a From tag (§8.1.1.3), `CSeq: 1 MESSAGE` matching the request method (§8.1.1.5), `Content-Type: text/plain` (RFC 3428), **`Content-Length` checked against the real body by arithmetic**, the body being byte-identical to the text the syslog record carries, and a round-trip back through the engine's own parser.

What that proves is the message is well-formed on the wire. What it does *not* prove is that any given handset renders an inbound MESSAGE to a human — untestable on host, and stated as its own limitation rather than blurred into the above.

## What this does NOT do

- **No SMTP.** `SmtpClient::sendAndWait()` blocks and its own contract forbids calling it from a SIP thread. Email belongs behind a future non-blocking `sendAsync`.
- **No webhook.** None exists; it would need an HTTP client on a path that must not block.
- **No retries, delivery receipts, retention, or PSAP notification.** None is required by the rule, and each adds a failure mode to a path whose whole value is that it cannot fail in a way that matters.
- **No dispatchable location conveyed with the call.** That is RAY BAUM'S Act §506 — a separate obligation, and one that lives with the trunk provider rather than here.
- **Nothing verified against a real carrier or a real handset.** The host suite proves the engine enqueues the notification after the call leg. It cannot prove a phone renders it.

`docs/FEATURE_ROADMAP.md`'s disclaimer stands and should stay: direct dialing plus notification is two legs of three, there is still no dispatchable location, and a 911 call competes for the single outbound anchor slot. Do not deploy this as anyone's only means of calling for help.

## Hardware

Untouched. Host suite only; no flash.


GlomarGadaffi:
Retargeted to `main` so CI actually runs. The workflow's `pull_request` trigger is filtered to `[main, master]` and its `push` trigger to `[main, master, 'feature/**']`, so a PR stacked on `feat/166-emergency-dialing` got **no checks at all** — which matters, since cppcheck is a hard gate in that job and is exactly what caught the `[mismatchingContainers]` finding on #237.

Until #241 merges this PR's diff shows both commits. It cleans up on its own once #241 lands; review #241 first.

GlomarGadaffi:
Vetted independently (same GitHub account, so a comment not a formal approval — verdict: **ship it**).

## What I checked
- Read `EmergencyNotifier.hpp`/`.cpp` in full, plus the `RequestsHandler::notifyEmergency()` wiring on all three exit paths of `routeEmergencyCall()` (routed, pool-exhausted, 503-no-trunk) — confirmed `notify()` genuinely fires on every one, always *after* the response is already enqueued, matching the 9.16(b)(2) ordering claim.
- Built from scratch in an isolated worktree: **833/833**, matches exactly.
- Independently mutation-tested the exact guard you flagged as newly-covered (`considered++ >= kMaxE911NotifyExts` inside `notify()`) — neutralized it, rebuilt, reran `TheNotifierBoundsTheListItselfEvenIfTheConfigDidNot`. Failed exactly as expected: 8 targets considered instead of 4. Reverted. Confirms the `CountingEnv`-driven test genuinely exercises the notifier's own bound independent of `setE911Config()`'s truncation, not just in the abstract.
- Confirmed `setE911Config()` and `loadE911()` both truncate to `kMaxE911NotifyExts` independently (write-time and load-time), so the notifier's own bound really is defence-in-depth against a foreign/future-build config blob, exactly as described — not dead code, not load-bearing for the public API either.
- Spot-checked the wire format by reading `buildNotifyMessage()` directly: Content-Length is computed from `text.size()` before assembly (not re-derived after), and `syncContentLength()` is still called afterward — consistent with the belt-and-suspenders pattern already used at every other SIP-message-construction site in this file (`RegisterBeeper.cpp` etc.), not a new inconsistency.

## Notes, not blockers
- `formatE911Notification()`'s 512-byte trim (`out.resize(512)`) truncates on a raw byte boundary. If `cfg.location`/`cfg.callback` ever contain multi-byte UTF-8 and the total happens to land past 512 mid-character, the tail could be a partial UTF-8 sequence. Extremely low likelihood (needs an unusually long operator-typed location/callback string) and low consequence (cosmetic, not a parsing hazard — the SIP MESSAGE body itself has no encoding requirement here) — not worth fixing, just noting.
- The 503 fallback's `Warning` text always says "no outbound trunk connected" even though `originateAnchorCall()` returning false could also mean the anchor-call pool is exhausted by unrelated concurrent calls. Doesn't affect correctness (503-not-404 is still right either way) — just a minor wording precision gap if it's ever worth distinguishing in a log/notification.

Same rigor as the rest of tonight's PRs and it shows — the compliance-relevant ordering is genuinely enforced in code (not just asserted in a comment), and chasing the mutation instead of deleting the guard was the right instinct.

GlomarGadaffi:
Re-vetted \`c9896ac\`. **Ship it — and good catch.**

## What I checked
- Traced all 9 \`refuse()\` call sites in \`originateAnchorCall()\` by hand (grep + read, not trusting the count): 8 of them are followed by \`return true\` (488 no codec, 400 bad RTP dest, 503×2 bridge-slot-busy, 503 makeCall declined, 503 media-bridge-failed, 503×2 session-pool-full) and all 8 route through the single \`refuse\` lambda, which now unconditionally clears \`*placedOut\` — one point of truth, not eight places to remember. The 9th (\`return false\`, "no anchor client connected") was already correctly handled pre-fix since \`routeEmergencyCall()\`'s own fallback branch already notifies \`routed=false\` for that case. The silent pool-exhaustion unwind (10th path, no \`refuse()\` call at all since it can't even get a message from the pool) also gets its own explicit clear.
- Checked the one path that stays \`placed=true\` without going through the synchronous 200-OK branch — the async \`makeCall\` dispatch path (real WAN trunk, TLS round-trip deferred off the SIP thread). It reports \`placed=true\` at the point \`asyncMakeCall()\` is genuinely dispatched, which is consistent with the synchronous branch's own semantics ("dispatched toward the trunk," not "PSTN confirmed") — not a gap this fix needs to cover, since a *later* async failure after dispatch is a different, out-of-scope problem (no notification-update mechanism exists or was promised).
- Independently mutated the `refuse` lambda's `*placedOut = false` line, rebuilt, reran `ACallRefusedForCapacityIsNeverReportedAsRouted`. It failed exactly as described, and the log output is about as convincing as a mutation-test failure gets — it shows, in order: `anchor(911): every anchor bridge slot busy for 200` immediately followed by `EMERGENCY: 911 dialed by ext 200 - ROUTED TO TRUNK`. Reverted.
- Hand-traced the UTF-8 boundary-trim loop against 2-byte, 3-byte, and straddling-sequence cases, plus the bounded worst case for malformed input (walks back to 0, never OOB, never infinite). Correct.
- Re-read the rewritten `PbxConfig.hpp` comment — it now states the #163→#166 history plainly instead of asserting something the code no longer does. Good resolution; I'd said "worth a one-line update," you were right that "worse than no comment, in a file about emergency dialing" was the sharper framing.
- Full suite rebuilt from scratch in a fresh worktree: **835/835**, matches exactly.

This was worth chasing rather than shipping with a weaker note. Nice work.

PR #245, opened 2026-09-15
fix(ci): run CI on every branch, not just main/master

GlomarGadaffi:
## Summary

`ci.yml`'s trigger block gated CI more narrowly than it needed to, in two independent ways:

- `pull_request.branches` was `[main, master]`, so a PR whose **base** is anything else — e.g. a stacked PR opened against an unmerged `feat/**` branch — got **zero CI**: no host tests, no cppcheck, nothing. `gh pr checks` just reports "no checks reported" instead of erroring, so this fails silently. This bit us for real tonight on #242.
- `push.branches` included `'feature/**'`, which (per `git blame`) has been there unchanged since the file was first added on 2026-06-01 and has **never matched a real branch** in this repo's history — every branch actually used is `feat/**`, `fix/**`, `docs/**`, or an ad-hoc name (`ci-perf-lints/`, `claude/`, `phase2-polish/`, ...). So a bare push to a feature branch (before any PR exists) also got no CI.

I checked `git blame` on the `on:` block before touching it — it's a single unmodified hunk from the file's initial commit, predating essentially all of the repo's current branch-naming convention, and no comment anywhere explains a deliberate reason for the narrower `pull_request` scope. For contrast, `ci-glolab.yml`'s `push`-to-`main`-only trigger *is* explicitly commented as deliberate (self-hosted runner, never run untrusted fork-PR code) — `ci.yml`'s block has no such comment, which supports this being drift rather than intent.

## Fix

Widen both `push.branches` and `pull_request.branches` to `[ '**' ]` (every branch), with a comment explaining why, instead of maintaining an allowlist of prefixes that will just drift again the next time someone invents a new naming pattern. Neither filter gained a `tags:` key, so this remains disjoint from `release.yml`'s tags-only trigger (a `push` matching only `branches` does not fire on tag pushes).

Net effect: `push` and `pull_request` now both fire for a branch that has an open PR against it, which double-runs CI for a normal PR — that costs runner minutes, not signal, and is a common/accepted tradeoff. (Confirmed below: this PR itself triggered two full runs, exactly as expected.)

## Test plan

- [x] Read `ci.yml` and `release.yml` in full before changing anything; confirmed the bug as described lives in `ci.yml`, not `release.yml` (`release.yml` is tags-only, no `pull_request` trigger at all).
- [x] `git blame`/`git log` on the `on:` block to confirm the narrow scope was drift, not a deliberate choice.
- [x] Diffed against every branch this repo has ever used (`git log --all --merges`) to confirm `'feature/**'` never matched anything, and to pick `'**'` over a prefix list.
- [x] Host build + full test suite green under WSL: `cmake -B build-wsl -S . -DCMAKE_BUILD_TYPE=Release && cmake --build build-wsl` then `ctest --test-dir build-wsl/tests --output-on-failure` — `sip_parser_tests` 100% passed.
- [x] **Live-verified the actual trigger fix**: this PR itself is proof — `gh pr checks 245` shows a full set of jobs (host tests, cppcheck, ESP-IDF matrix, deployment audit, CodeQL) actually scheduled, across two separate runs (one `push`-triggered, one `pull_request`-triggered — [run 35024104191](https://github.com/GlomarGadaffi/pocket-dial/actions/runs/35024104191), [run 35024134935](https://github.com/GlomarGadaffi/pocket-dial/actions/runs/35024134935)). Before this fix, a branch/PR combination like this one would have shown "no checks reported."


GlomarGadaffi:
Vetted (same account, so a comment not a formal approval — verdict: **ship it**).

Verified the git blame claim myself: the whole `on:` block is byte-identical back to `8be61da5` (2026-06-01, "docs: refactor README and add comprehensive post-refactor documentation suite") — confirms this predates the repo's real branch naming, `'feature/**'` never matched anything real. Good correction of my original brief too — I'd pointed at release.yml, which is tags-only and was never the actual gap; ci.yml (host tests, cppcheck, the ESP-IDF matrix) was.

`branches: ['**']` over a prefix allowlist is the right call — an allowlist just moves the same failure mode to the next time someone invents a new prefix, and this repo's history already has four or five different ones. Keeping it disjoint from release.yml's tags-only trigger (no `tags:` key added here) is correct and worth having said explicitly in the comment.

The self-verification via the PR's own two live CI runs (push + pull_request both firing, jobs actually scheduled instead of "no checks reported") is exactly the right proof for a workflow-trigger fix — better than anything I could confirm by reading the YAML alone. Nice, clean fix.

PR #247, opened 2026-09-15
E.164 follow-ups from #237 review: stale stored rendering + table-wide false-positive scope

### LINKED ISSUE #243: E.164 follow-ups from #237's review: stale stored DID rendering, and the false-positive scope is table-wide
Two non-blocking notes from the review of #237 (E.164 normalization). Both land after #237 merges — they could not go on #166's branches because `E164.hpp` does not exist there yet.

## 1. `setMapping()` keeps the OLD stored rendering on update

`DidMapping::findIndex()` matches on E.164 equivalence, so `setMapping()` with a *different rendering* of an existing DID correctly updates in place rather than adding an ambiguous second row. But it only assigns `.extension` — `.did` keeps whatever spelling was stored first.

Effect: an operator who re-enters a DID as `+12025550123` over a stored `(202) 555-0123` sees the old form in `list()` and on the dashboard, with no way to correct it short of delete-and-re-add.

Fix is one line in the update branch:

```cpp
_entries[idx].did = did;        // adopt the new rendering
_entries[idx].extension = extension;
```

Worth a test asserting the stored rendering follows the latest write.

## 2. The accepted false positive is table-wide, not pairwise

`E164.hpp` documents that `+442025550123` and `2025550123` compare equal, and argues the collision needs a deliberate coincidence. That reasoning is right but scoped too narrowly: `extensionForDid()` scans **every row**, so one inbound DID gets N chances to collide, not one.

What actually bounds the risk is `POCKETDIAL_MAX_DID_MAPPINGS` (8) plus the fact that each row is a number one operator deliberately typed for one carrier — **not** anything in the comparison itself. A deployment that raised that cap substantially would be raising this risk with it.

Documentation-only: add the scope note to `e164SameNumber()`'s boundary comment so the bound that does the work is named where someone tuning `POCKETDIAL_MAX_DID_MAPPINGS` will see it.

Found by PD-Sonnet-Orchestra's vet of #237.

GlomarGadaffi:
## Summary

Two non-blocking follow-ups filed after #237 (E.164 DID normalization) merged.

**1. `DidMapping::setMapping()` kept the OLD stored rendering on update.**

`findIndex()` matches on E.164 equivalence (#165), so re-setting an existing DID with a different rendering correctly updates in place instead of adding an ambiguous second row -- but the update branch only assigned `.extension`, leaving `.did` at whatever spelling was stored first. An operator who re-entered a DID as `+12025550123` over a stored `(202) 555-0123` saw the old form in `list()`/the dashboard, with no way to correct it short of delete-and-re-add.

Fixed: the update branch now adopts both the new rendering and the new extension. Added `DidMappingTest.UpdateInPlaceAdoptsTheNewRendering` pinning that the latest write's rendering is what `list()` reports afterward, while the DID stays findable under either rendering.

**2. The accepted E.164 false positive is table-wide, not pairwise.**

`e164SameNumber()` itself only ever judges one pair, and the boundary comment described the false positive as if reaching it were a single coincidence between two specific renderings. That's too narrow: `DidMapping::findIndex()` calls it once per **stored row**, so one inbound DID gets as many chances to collide as there are rows in the table, not one.

Sharpened the comment to name `POCKETDIAL_MAX_DID_MAPPINGS` -- not anything in the comparison itself -- as what actually bounds that exposure, so a deployment raising the cap sees the tradeoff it's tuning. Documentation-only.

## Test plan

Full host suite built and run under WSL against this worktree:

```
[==========] 876 tests from 121 test suites ran. (30109 ms total)
[  PASSED  ] 876 tests.
```

Includes the new `DidMappingTest.UpdateInPlaceAdoptsTheNewRendering` test and confirms no regressions elsewhere (`DidMapping_test.cpp`, `E164_test.cpp`, and the full suite).

Closes #243

Generated with Claude Code


GlomarGadaffi:
Vetted — verdict: **ship it**. Both fixes are exactly what was asked, minimally scoped, and the doc rewrite in E164.hpp correctly generalizes the false-positive risk from pairwise to table-wide (bounded by POCKETDIAL_MAX_DID_MAPPINGS), matching what I'd flagged during tonight's original #237 review. The new test pins the actual behavior (list() reflects the latest rendering, both spellings still resolve) rather than just re-asserting the extension update. Clean.

GlomarGadaffi:
Independent second review (Fable session, did not author). Verdict: **approve.** All checks green at the time of writing.

Checked myself:

- **Validation order.** `setMapping()` runs the "DID required / Field too long / control character" checks *before* `findIndex()`, so adopting the new `did` rendering on the update path can never store an unvalidated string. Persist happens after, same as the extension update always did.
- **Non-telephone DIDs unaffected.** Exact-string matching for hand-edited non-numeric entries is untouched; adopting the caller's rendering on those is a no-op since exact match means identical text.
- **Test pins the behaviour, not the mechanism.** `list()[0].did` follows the latest write and both renderings still resolve afterward — that is the operator-visible contract from #243, item 1.
- **E164.hpp scope note** correctly names `POCKETDIAL_MAX_DID_MAPPINGS` as the real bound and says the risk scales with the table, which is what #243 item 2 asked for and where someone raising the cap will read it.

Nothing to add. Minimal, scoped to the issue.

Issue #199, opened 2026-09-14
RFC compliance map: 85 RFCs assessed — two root causes, and what each gap actually costs


GlomarGadaffi:
Tracking issue. **85 RFCs assessed** against the source, with the load-bearing claims put through an adversarial verification pass (4 of 19 checked claims were refuted — all in the direction of *"more partial than claimed"*, so treat the statuses below as generous rather than harsh).

This is not a conformance-badge exercise. The point is to make explicit which gaps cost something today, which block #194, and which are defensibly skipped — so skipping is a decision rather than an accident.

---

## The two root causes

Most Tier 1 entries below are downstream of exactly two structural gaps. Fixing these two would move a dozen rows at once.

### 1. There is no SIP *server* transaction layer

`TransactionLayer.hpp:14` scopes itself to the "§17 INVITE **client** transaction", and `TransactionLayer::classify()` (`TransactionLayer.cpp:10-13`) returns `None` for responses and for every non-INVITE request. Consequences, all verified by grep:

- **No §13.3.1.4 2xx-retransmit-until-ACK**, no §17.2.1 Timers G/H/I, no §17.2.2 Timer J.
- **No non-INVITE *client* transaction either** (Timer E/F), so every PBX-originated `BYE`, `NOTIFY`, `REFER`-NOTIFY and `CANCEL` is **fire-and-forget, sent exactly once**.

The codebase is candid about it — the recurring comment idiom is *"pool exhausted: drop, peer retransmits"* — but that only delegates reliability for messages the **peer** retransmits. Nothing retransmits ours. One lost UDP packet leaks state permanently: a dropped BYE leaves a handset showing a live call forever (park retrieve, pickup, transfer splice, session-timer reap, register-beep teardown all end this way), and a dropped NOTIFY freezes a BLF lamp until that extension's dialog state next changes.

Related: #198 (session-timer reap) and #197 (blind transfer) both ride this path.

### 2. No capability headers are ever sent

There is no `Allow:`, `Supported:`, `Require:` or `Accept:` header anywhere in `src/`. The only capability header in the tree is `Allow-Events: dialog` (`BlfSubscriptions.cpp:154`).

This one omission silently defeats **three features the README advertises as shipped**, because a conformant phone gates each on a capability it never sees:

| Feature | Gate | Result |
| :-- | :-- | :-- |
| **UPDATE** (3311 §5.1) | `Allow: UPDATE` | `onUpdate` is dead code against well-behaved phones |
[...]

Issue #131, opened 2026-09-04
Attended transfer: REFER with Replaces does not splice B and C's dialogs


GlomarGadaffi:
Found by the ported \`.smoke/office_smoke.py\` while baselining #127/#128 (scenario "Attended transfer": \`B got C's SDP=False, C got B's SDP=False, A dropped from both=False\`). Not yet compared against drawbridge — unlike #127/#128, no prior confirmation that drawbridge passes this scenario; needs that check first.

**Repro (host build, loopback):** A calls B; B answers. A separately calls C; C answers. A sends REFER on the A-B dialog with \`Replaces=<A-C call-id>\` (attended transfer). Expected per RFC 3891/5589: B and C's dialogs get spliced together (each gets a re-INVITE carrying the other's SDP), and A drops out of both legs.

**Observed:** none of that happens — B never receives C's SDP, C never receives B's SDP, and A is not dropped from either leg.

**Where to look:** \`RequestsHandler::onRefer\` — issue #128's body notes this repo's REFER handling is "blind-only," which may mean the \`Replaces\` header on the REFER's Refer-To is not parsed/handled at all yet, i.e. this could be a missing feature rather than a regression. Confirm against drawbridge (or the office_smoke.py scenario's own docstring, scenario 5) before assuming there's prior working code to restore. \`.smoke/office_smoke.py\` scenario \`scenario_attended_transfer\` is the regression test once implemented/fixed.

Filed while working #127/#128 in an autonomous /loop session (Claude Code) — not yet investigated beyond the office_smoke.py baseline run.

GlomarGadaffi:
Correction to the original filing — I guessed this might be an unimplemented feature; it's not. drawbridge has attended transfer: commit e594914 "feat(sip): Phase C-2 — attended transfer via REFER + Replaces (RFC 3515/3891)". Confirmed scope: `src/SIP/RequestsHandler.{cpp,hpp}`, `src/SIP/Session.hpp` (14 lines: `_remoteSdp`/`getRemoteSdp`/`setRemoteSdp`, `_isTransferBridge`/`isTransferBridge`/`setTransferBridge`), plus its own test file. No MediaBridge/RtpSender involvement.

[...]

PR #248, opened 2026-09-15
fix(sip): stop conflating Supported: replaces with BLF-key INVITE pickup

### LINKED ISSUE #229: Supported: replaces is an over-claim — there is no INVITE-with-Replaces handling, so BLF-key pickup still cannot work
Split out of #199's root cause 2, where the RFC 3891 row reads:

> | **Replaces** (3891 §4) | `Supported: replaces` | BLF-key call pickup silently fails |

That framing says the gate is the missing header. It isn't — or rather, it is only the smaller half. #200 added `replaces` to `Supported` and #226 put that advertisement on the paths phones actually read. **BLF-key call pickup still will not work**, because the thing the header advertises is not implemented.

Two independent reads of the tree (during #226) landed on the same finding, so this is not a one-pass impression.

## What is actually implemented

`Supported: replaces` tells a phone it may send **an INVITE carrying a `Replaces:` header** (RFC 3891 §3: *"A UA that supports this specification MUST... accept an INVITE with a Replaces header"*). That is the mechanism behind "grab this ringing call from the BLF key" and phone-native call steal.

There is **no handling of a `Replaces` header on an INVITE anywhere in `src/`.** `grep -rn Replaces src/` returns only:
- the `?Replaces=` **URI parameter** inside a `Refer-To` header, parsed in `onRefer()`
- comments and unrelated English uses of the word "replaces"

`onInvite()` has no `Replaces` branch at all, so such an INVITE is treated as a brand-new call: the PBX rings the target again rather than replacing the named dialog. From the phone's side the pickup key appears to do nothing useful.

So what the PBX supports is **REFER with `?Replaces=`** (attended transfer, issue #131) — a different RFC 3891 use, reached by a different method.

## Two smaller gaps in the part that does exist

Line numbers are against `feat/sip-transaction-layer` @ `518fb66` (what `main` becomes when #226 merges) — anchor on the symbol names, since #199 taught us line refs here go stale fast.

**1. Dialog matching is Call-ID only.** `RequestsHandler.cpp:4470`:

```cpp
if (auto consult = getSession("Call-ID: " + replacesCallIdBare);
```

The from-tag and to-tag are parsed out of the URI and then deliberately discarded — `RequestsHandler.cpp:4395-4409` truncates at the first `;`, both before and (after URL-decoding) again for a `%3B`-encoded one:

```cpp
std::string repVal = uriParams.substr(rp + 9);
size_t semi = repVal.find_first_of(";&>\r");
replacesCallIdBare = siphdr::urlDecode(
    (semi == std::string::npos) ? repVal : repVal.substr(0, semi));
```

RFC 3891 §3 specifies matching on the full dialog triple (Call-ID + to-tag + from-tag) precisely because a Call-ID can name more than one dialog — forked requests being the obvious case. Call-ID-only matching is the dialog-confusion hazard the section was written to prevent.

Worth saying clearly: **this is not currently a security hole.** `isDialogSourceAuthorized()` gates the operation on the requester actually being a party to the dialog, and there is a second check that the transferor's number appears on both dialogs. The truncation is also deliberately hardened against a `%0D%0A`-smuggled CRLF (a control-character guard follows immediately). The gap is correctness under forking, not authorization.

**2. `early-only` is not parsed or honoured** (RFC 3891 §4). No occurrence anywhere. A phone sending `Replaces: ...;early-only` expects the replacement to be refused if the named dialog has already been answered; today the parameter would simply be ignored — though since the header is not handled at all, this only becomes live once it is.

## Why this is filed rather than fixed in #226

#200 decided the `replaces` claim was honest enough on the strength of the REFER path, with reasoning in the comment above `kSupportedOptionTags`. #226 did not relitigate that — it is a scope call, not a bug, and reversing it inside a transaction-layer PR would have been the wrong place. But the decision deserves to be a recorded, revisitable one rather than an accident, and #199's own table should not be left implying that a header fixed the feature.

## Options, roughly in cost order

1. **Do nothing, correct the documentation.** Amend #199's RFC 3891 row so it reads as "REFER `?Replaces=` only; INVITE-with-Replaces absent", and leave `Supported: replaces` as is on the grounds that the REFER path is real and widely exercised. Cheapest, and arguably already the status quo — it just stops the tracker overstating what a header bought.
2. **Drop `replaces` from `Supported`** until the INVITE path exists. Most literally honest, and consistent with how `timer` is handled — but it would *remove* a capability phones use today to offer attended transfer via REFER, so it likely makes things worse in practice. Mentioned for completeness; I would not recommend it.
3. **Implement INVITE-with-Replaces.** Parse the header, match the full dialog triple, honour `early-only`, and on a match answer the replacing INVITE and BYE the replaced dialog. This is what actually delivers "BLF-key call pickup" and phone-native call steal. It also wants the tag-matching fix in (1) above applied to the REFER path so both use one matcher.

My read: (1) now, (3) when BLF pickup is genuinely wanted — and note (3) pairs naturally with RFC 4235's missing dialog identifiers, which #199 flags separately ("a watcher can never construct the Replaces needed to pick the call up from the key"). Those two are the same feature and are probably one piece of work: **the watcher cannot build the header today even if the PBX could read it.**

Found while implementing #199's two root causes (#226).

GlomarGadaffi:
## Summary

Closes #229.

`Supported: replaces` covers two independent RFC 3891 mechanisms that
happen to share one option tag:

- REFER whose Refer-To carries a `?Replaces=` URI parameter (attended
  transfer, issue #131) — **implemented**, `onRefer()` genuinely splices
  the two dialogs.
- An INVITE that itself carries a `Replaces:` header naming an existing
  dialog by Call-ID + both tags (§3) — the actual BLF-key "grab this
  ringing call" / phone-native call-steal mechanism — **not implemented
  anywhere in `src/`**. `onInvite()` has no Replaces branch, so such an
  INVITE just rings its own target as an ordinary new call.

#199's RFC compliance map read as though the header alone bought BLF-key
pickup. It doesn't — that's the smaller half of what the tag means. This
was filed separately (#229) rather than fixed inside #226 because it's a
scope call, not a bug in what #226 shipped.

## What I did

Per the issue's own cost-ordered options list, I took option 1 ("do
nothing [behaviorally], correct the documentation") rather than option 3
(implement full INVITE-with-Replaces). The issue's author explicitly
recommends option 1 now and option 3 only "when BLF pickup is genuinely
wanted" — full support needs whole-dialog Call-ID+tag matching and
`early-only` handling to be RFC-correct, which is real, separate work I
did not want to bundle into a doc-accuracy fix.

Concretely:
- Corrected every comment in `RequestsHandler.cpp` that implied
  `Supported: replaces` alone enables BLF-key pickup (it doesn't — it
  backs the REFER path only), while explicitly keeping the tag advertised
  (dropping it would break the REFER attended-transfer path that is real
  and exercised, for no gain).
- Corrected the matching comments in `CapabilityHeaders_test.cpp` that
  used the same "Replaces pickup" language.
- Added `ReplacesInvite_test.cpp` as a trip-wire: it pins that an INVITE
  carrying a `Replaces:` header is currently a complete no-op (the named
  dialog is untouched — no BYE, no re-INVITE, no state change — and the
  INVITE is dispatched as an ordinary new call). If INVITE-with-Replaces
  is implemented later, this test has to be edited on purpose rather than
  silently keep passing for the wrong reason.

No behavioral change. `Supported: replaces` is still advertised, exactly
as before, because the REFER-based use it backs is real.

## Test plan

- [x] Full host suite (WSL, `build-wsl/tests/sip_parser_tests`): **876/876 passing**
- [x] New `ReplacesInvite.*` test + full `CapabilityHeaders.*` suite verified together


GlomarGadaffi:
Vetted — verdict: **ship it**. Genuinely good call reading the actual issue rather than the framing I gave in the dispatch brief — #229 had already worked through "implement vs. withdraw" and picked a third option (keep the tag, since REFER ?Replaces= / #131 is real and withdrawing would break it; document and pin the actual gap instead), and that's clearly the right scope call: full INVITE-with-Replaces needs whole-dialog Call-ID+both-tag matching and early-only handling to be RFC-correct, not a quick header read, so this isn't a case of scoping down to avoid work.

The regression test is real end-to-end coverage (establishes an actual A-B dialog through handle(), sends the Replaces INVITE to a separate extension D so a busy-target ambiguity can't confound the assertion), not a unit-level stub. No production behavior changed, confirmed by reading the diff, not just the summary.

GlomarGadaffi:
Independent second review (Fable session, did not author). Verdict: **approve, gated on the pending Host Build check** — it was still `pending` at the time of writing; the firmware matrix is green.

Checked myself:

- **The claim "no handler for INVITE-with-Replaces" is true on main.** Only `onRefer()`'s `?Replaces=` URI-parameter parse exists; `onInvite()` has no `Replaces` branch. The rewritten comment above `kSupportedOptionTags` now states exactly what the tag backs (REFER attended transfer, #131) and what it does not.
- **The trip-wire test is end-to-end.** It establishes a real A→B dialog through `handle()` with a relayed 200 OK, sends the Replaces INVITE from C toward a *fourth* extension D, and asserts on captured wire bytes: no BYE to A or B, no re-INVITE to B, A→B still `Connected` with dest 501, and a fresh fork to D in `Invited`. Sending it toward D rather than B is the right choice — it removes the busy-target ambiguity so the "treated as a brand-new call" assertion is unambiguous.
- **No behavioural change**, confirmed from the diff: the only non-comment source edit is the CMake registration of the new test.

Two things worth recording, neither blocking:

1. **This documents a deliberate RFC deviation, and the comment should say so in those words.** RFC 3891 §3 says a UA advertising `replaces` MUST accept an INVITE with Replaces. Keeping the tag advertised while not implementing that half is the scope call #229 recommended, and it is the right one for the REFER path's sake, but the comment currently frames it as "does not cover" rather than "we knowingly advertise more than we implement". One sentence naming it as a known deviation would stop the next reader from re-filing #229. Fine as a follow-up edit or a one-line tweak here.
2. **The PR description ends with a Claude session URL.** desmo's standing rule tonight is no session URLs in commits or PR bodies. The squash commit will not carry the body, but the description should still be edited before merge.

Issue #228, opened 2026-09-14
forceDisconnect() (/api/kill) tears down a call server-side only — sends no BYE to either party


GlomarGadaffi:
## Problem

`RequestsHandler::forceDisconnect(const std::string& extension)` (`RequestsHandler.cpp:5538-5581`), the admin action behind `/api/kill` (`sendApiKill`), tears down a live call **entirely server-side** — it releases the client, erases the `Session` from `_sessions`/`_sessionPool`, and stops the PBX's own RTP if it owned any (`_rtpSender.stop(callID)`) — but it never builds or sends a `BYE` to either party.

```cpp
void RequestsHandler::forceDisconnect(const std::string& extension)
{
	...
	for (auto& client : _clientPool)
	{
		if (client->getNumber() == extension)
		{
			client->release();
			break;
		}
	}
	// Also remove any sessions involving this extension
	for (auto it = _sessions.begin(); it != _sessions.end(); )
	{
		...
		it = _sessions.erase(it);
		...
	}
	...
}
```

No `_outbox`/`_asyncOutbox` push, no `BYE` construction, anywhere in the function.

## Impact

For an ordinary ext-to-ext call (peer-to-peer media, per this engine's governing constraint), the PBX force-disconnecting extension A while A is on a call with B:

- **A's session state is gone from the PBX**, but A's own handset was never told the call ended — if A hung up locally it's fine, but nothing about this function tells A's phone to hang up.
- **B's handset is never notified at all.** From B's perspective the call is still live — B keeps holding a call the PBX has completely forgotten about. B's RTP stream to A continues indefinitely (this is P2P media, the PBX was never in that path to stop it either).

The admin dashboard's "kill" action is presented as ending a call; it doesn't. The phones only discover the call is over when one side eventually hangs up locally and its own BYE arrives at a PBX with no matching session (`481 Call/Transaction Does Not Exist`), or never — if neither party notices, the call sits open until each phone's own idle/registration-refresh behavior does something about it, which could be a long time.

## Where this was found

Found by PD-Opus (issue #199 root-cause-1 work) while attempting to use `forceDisconnect()` as a server-BYE trigger for a transaction-layer test — it doesn't send one, so the test didn't work.

## Suggested fix

[...]

Issue #72, opened 2026-06-03
Add zero-copy SipStatus parser; route SIP responses by numeric status code


GlomarGadaffi:
## Summary
Introduce a lightweight, zero-copy, ROM-friendly SIP status-line parser (`SipStatus`) and use it to route incoming responses by numeric status code, replacing the brittle exact status-line string matching.

## What's added
- `src/SIP/SipStatus.{hpp,cpp}`:
  - `parseSipStatusLine(std::string_view) -> std::optional<SipStatusInfo>` (pure, `noexcept`).
  - `SipStatusInfo { uint16_t code; SipStatusClass klass; bool softFail; }` — a trivially-copyable POD with no `string_view` members, so **zero lifetime concerns**.
  - `constexpr` ROM table of standard statuses (code → reason, soft/hard fail classification).
- `SipMessage`: `parse()` populates `std::optional<SipStatusInfo> _statusInfo`, exposed via `getStatusInfo()`.
- `RequestsHandler::handle()`: dispatches on `status->code` (100 / 180 / 200 / 480 / 486 / 487); requests and non-status lines fall back to the method token.

## Build / test wiring
- Registered `SipStatus.cpp` in `main/CMakeLists.txt` (firmware) and in the host `sip_parser_tests` target.
- Added `tests/SipStatus_test.cpp`.

## Verification
- ESP-IDF firmware build (esp32s3) green; app binary fits with ~22% free.
- MSVC C++17 host build green; **10/10** host tests pass.
- Built and run on the aarch64 Raspberry Pi target.

Issue #198, opened 2026-09-14
RFC 4028: PBX never generates a session refresh (it is never the refresher)


GlomarGadaffi:
## Summary

The PBX arms an RFC 4028 session timer from a callee's `200 OK`, **never refreshes it**, and BYEs both legs when it expires. A phone that puts `Session-Expires` in its answer unilaterally — common Yealink/Grandstream behaviour — therefore gets a healthy call torn down mid-conversation, with no cause visible to either party.

`docs/FEATURE_ROADMAP.md:63` describes the support accurately as **"Passive"**. `README.md:96` lists "RFC 4028 session timers" as a shipped feature without that caveat.

## The path

1. **Arm.** `RequestsHandler::armSessionTimer` (`src/SIP/RequestsHandler.cpp:5601-5610`) reads `Session-Expires` from the callee's `200 OK` and arms unconditionally when present:

```cpp
uint32_t secs = ok200->getSessionExpiresSecs();
if (secs == 0) return;
auto ref = ok200->getSessionExpiresRefresher();
bool weRefresh = (ref == "uas");
session->armSessionTimer(secs, weRefresh, std::chrono::steady_clock::now());
```

Called once at `:3345`, on the connect path. Re-armed on re-INVITE (`:5455-5460`) and UPDATE (`:5507-5512`, `:5542-5547`).

2. **Never refresh.** Nothing in the codebase ever sends a refreshing re-INVITE or UPDATE. `Session::getNextRefresh()` (`Session.hpp:129`) and `isRefresher()` have **zero consumers** that generate one, so the `weRefresh` flag computed above is dead.

3. **Reap.** `sweepSessionTimers` (`:5612-5660`), driven from `tick()`, BYEs **both legs** on expiry.

So whenever the phones themselves do not refresh, the timer runs to expiry and the PBX hangs up a working call.

## Why phone-to-phone usually survives, and when it does not

For an ordinary extension-to-extension call both phones typically refresh on their own cadence, and the relayed re-INVITE/UPDATE re-arms the timer at `:5455`/`:5509`. That is why this has not been loud.

It stops being benign when:

- the phone designates the **PBX** as refresher (nothing will ever refresh), or
- the phone offers an abusively short `Session-Expires` — the PBX cannot negotiate it upward because it never returns `422` (see below), so it simply BYEs on that cadence, or
- the call is anchored (`555`/trunk/`888`), where there is no peer re-INVITE to re-arm from.

[...]

PR #249, opened 2026-09-15
fix(sip): record 777/888/555 answer-leg tags so forceDisconnect() can BYE them

### LINKED ISSUE #232: forceDisconnect(): killed handset on a 777/888 leg gets no BYE (answer paths never record their To-tag on the Session)
## Problem

Follow-up to #228 / PR #231. `forceDisconnect()` now BYEs both handset legs of an ordinary call, but on a **virtual-extension leg** (777 echo, 888 conference) the killed handset still gets no BYE and keeps the call up until it hangs up locally into a 481.

Root cause: those answer paths generate a To-tag for their own 200 OK and then drop it — nothing records it on the `Session`, so no server-side teardown can build a dialog-correct BYE toward the handset afterwards. `Session::getDialogFrom()/getDialogTo()` stay empty on these legs, which is exactly the #72 guard that suppresses the BYE in `forceDisconnect()` and `sweepSessionTimers()`.

Sites on `main` @ 65c64a7 (`src/SIP/RequestsHandler.cpp`):
- 888 conference: `const std::string toTag = IDGen::GenerateID(9);` at `:2154`, used by `buildOkWithSdp()` at `:2161`, never stored.
- 777 echo: `const std::string toTag = IDGen::GenerateID(9);` at `:2669`, `buildOkWithSdp()` at `:2674`, never stored.
- 555 anchor: `:377-389` already reads `session->getLocalTag()`, so that leg has the tag; only `setDialogHeaders()` is missing there.
- `ParkOrbit.cpp:32` generates its own tag too (its doc comment says it handles dialog headers itself — verify rather than assume).

## Suggested fix

At each answer site, after the 200 OK is built: `session->setLocalTag(toTag)` and `session->setDialogHeaders("From: " + <request From>, "To: " + <request To> + ";tag=" + toTag)` (header-name-prefixed, the convention every other `setDialogHeaders()` caller follows — see the comment at `:4939`). Then in `forceDisconnect()` drop the `destIsVirtual` exclusion for the **src** BYE only (the stand-in dest must still never be BYEd — it carries the caller's own address). `sweepSessionTimers()` deliberately skips these legs for a different reason (#198) and should stay as is.

Test: extend `tests/ForceDisconnect_test.cpp`'s `EchoLegFreesTheServerTransactionAndNeverByesTheStandInPeer` to expect exactly one BYE to the caller with `To` carrying the caller's tag and `From` carrying the stored `toTag`.

Found while scoping #228; deliberately left out of #231 because no far-end phone exists on these legs, so the original bug (peer never told) does not apply.

GlomarGadaffi:
## Problem

Follow-up to #228 / PR #231. `forceDisconnect()` now BYEs both handset legs of an ordinary call, but on a virtual-extension leg (777 echo, 888 conference, 555 anchor) the killed handset still got no BYE and kept the call up until it hung up locally into a 481.

## Root cause

The bug is **not** in `forceDisconnect()`'s teardown logic — its #72 guard (refuse to build a BYE from an empty From/To) and its `destIsVirtual` exclusion (never BYE the stand-in dest, which carries the caller's own address) are already scoped correctly, one check per leg.

The actual root cause is upstream: on these three answer paths the PBX is the UAS and authors its own 200 OK. Each one generates a To-tag for that 200 OK and then drops it — nothing ever calls `Session::setLocalTag()`/`setDialogHeaders()`, so `getDialogFrom()`/`getDialogTo()` stay empty forever and no server-initiated teardown (`forceDisconnect()`, `sweepSessionTimers()`) can ever build a dialog-correct BYE toward the handset. `ParkOrbit.cpp` was already doing this correctly at both its park and retrieve sites — verified rather than assumed, per the issue's own note.

## Fix

Record the tag the same way `ParkOrbit`/`CallPickup` do at answer time, at all three affected sites in `src/SIP/RequestsHandler.cpp`:
- the 777 echo loopback (`onInvite`'s `"777"` branch)
- the 888 conference join (`onConferenceInvite`)
- the 555 anchor's async `Answered` event callback, which already reused the Ringing leg's tag via `getLocalTag()` but never stored the dialog headers

The synchronous loopback anchor branch (test-only `LoopbackAnchorClient`, used by `AnchorRouting_test.cpp`) is untouched — it wasn't cited in #232, and it's not a path any real deployment runs.

## Testing

Extended `ForceDisconnect.EchoLegFreesTheServerTransactionAndNeverByesTheStandInPeer` to prove the fix rather than just the absence of a bug: the stand-in dest (which shares the caller's own address) must still never get a BYE, but the caller's own leg now does, with the exact tags the echo leg's own 200 OK used.

Sanity-checked the test is a real regression test, not a vacuous one: reverted the production change only, rebuilt, and confirmed the test fails (`toCaller.size()` is 0, not 1) — then restored the fix and confirmed it passes again.

Full host suite: **875/875 passing**.

Closes #232

Generated with Claude Code


GlomarGadaffi:
Vetted (same account, comment not a formal approval — verdict: **ship it**).

Traced the fix against the codebase's existing setDialogHeaders() convention (ParkOrbit.cpp, CallPickup.cpp, the ordinary onOk() path at :7480) rather than taking the "same convention" claim on trust: From = caller's original From-with-tag, To = the answering side's own newly-minted To-tag, matches exactly at all three new call sites. The test's From/To-swap assertion for the outgoing BYE (PBX's own tag becomes From, caller's original tag becomes To) is correct SIP dialog mechanics for a UAS-originated teardown, and it's a real assertion against captured wire bytes, not a comment claim.

Genuinely good root-cause work — the issue's own framing pointed at forceDisconnect() itself, and the actual bug is three upstream answer-construction sites that never recorded dialog state in the first place. The revert-rebuild-confirm-fails-restore sanity check on your own test is exactly right or a "fix" here.

GlomarGadaffi:
Independent second review (Fable session, did not author). Verdict: **approve, gated on the pending host-build and firmware-matrix checks finishing green** — at the time of writing the ESP-IDF esp32s3/wifi job and both Host Build runs were still `pending`, so this is not a merge signal until they land.

Checked myself rather than from the first review:

- **Header convention.** `SipMessage::getFrom()/getTo()` return the *whole header line* (`From: <sip:...>;tag=...`), so `std::string(data->getFrom())` is already header-name-prefixed. That matches the ordinary `onOk()` capture (`setDialogHeaders(ok200->getFrom(), ok200->getTo())`) and ParkOrbit/CallPickup exactly. The issue text's `"From: " + <request From>` wording would have double-prefixed; the PR did the right thing.
- **forceDisconnect() untouched, correctly.** The issue suggested dropping the `destIsVirtual` exclusion for the src BYE, but the src BYE was never excluded — it was only gated on non-empty dialog headers. Recording the headers is the whole fix; the dest stand-in stays silent. Verified in the branch's `forceDisconnect()`.
- **BYE direction.** src BYE is built with `(dTo, dFrom)`: From = the PBX's own To-tag, To = the caller's original tag. Correct for a UAS-originated teardown, and the test asserts both against captured wire bytes.
- **#198 preserved.** `sweepSessionTimers()` skips 777/888/555 legs by dest number with a `continue` *before* the dialog-header guard, so recording headers does not re-enable the spurious reaper BYE #198 was about.
- **555 site.** `toTag` there is `session->getLocalTag()` reused from the 180, same lambda scope, so the new `setDialogHeaders()` call is consistent with the tag the 200 actually carried.

Non-blocking gap: only the 777 echo test was extended. The 888 conference and 555 anchor sites got identical edits but no assertion; a follow-up test for at least 888 would stop a future refactor from silently dropping one of the three. Not a reason to hold this PR.

PR #250, opened 2026-09-15
SIP trunk part 1: outbound ITSP dialog, and the raw RTP relay it needs (#164)

GlomarGadaffi:
Part 1 of N for #164. Outbound only.

## Read this first: SipTrunk is compiled but NOT in the binary

Nothing constructs a `SipTrunk` yet, and the IDF link uses `--gc-sections`, so the class is garbage-collected out of the firmware image. This is correct for a slice with no engine wiring, but it means **this PR is reviewable design plus proven primitives, not a working trunk.**

Verified rather than assumed — the object file carries every symbol and string, the binary carries none of them:

```
SipTrunk.cpp.obj   826 KB   "Trunk: INVITE" present, 4 SipTrunk symbols
SipServer.bin               "Trunk: INVITE" ABSENT
SipServer.bin               "no sink of any kind" PRESENT   <- RtpReceiver, has real callers
```

The `RtpReceiver` half of this PR **is** in the binary, because it has callers.

This makes the third instance in this repo of built-with-zero-callers code, after `SipRegistrationClient` (#214) and the whole RFC 4733 receive path before #199. The binary string-grep is the cheap check that tells "shipped" apart from "compiles".

## Why this is not an AnchorClient

#164's own "Shape of the work" proposes building the trunk behind the existing `AnchorClient`/`MediaBridge` extension point. That cannot work, and the issue text has been corrected in a comment.

`AnchorClient` is a **PCM-16 audio** contract. RFC 4733 telephone-event packets have no representation in `writeAudio()` or `AudioRxCallback` — there is no method on that interface by which a DTMF digit could cross it. A trunk routed through it would decode the carrier's telephone-event packets to PCM16 and re-encode them, destroying them: no IVR, no calling card, no conference bridge PIN.

So signaling lives in the engine as a dialog reusing `TransactionLayer`, and media is a separate `RtpReceiver`/`RtpSender` pair in raw relay. The reasoning is restated at the top of `SipTrunk.hpp` where the next person to try it will read it.

## The prerequisite that turned out not to exist

This work was scoped behind "gate the star-code parser so it only claims digits on legs the board terminates" — the concern being that `RtpReceiver` diverts every non-PT-0 packet into `dispatchDtmf()` and so into the local feature-code parser, meaning a caller pressing keys at a carrier IVR would trip a local star code.

That call path is real but **inert unless armed**:

- `RtpReceiver.hpp` — `std::atomic<uint8_t> _dtmfPt{kDtmfPayloadTypeUnset};`
- `RtpReceiver.cpp` — `dispatchDtmf()` returns `false` at its first guard when unset
- The only production caller of `setDtmfPayloadType()` is `MediaBridge.cpp`, and `MediaBridge` is `init()`ed only for anchor legs (`RequestsHandler.cpp`) and conference legs (`ConferenceRoom.cpp`) — both board-terminated

A trunk receiver that never arms DTMF cannot reach the parser. **No gating change was required and none was made.**

## The gap that was actually there — the inverse

Because a trunk leg must not decode DTMF locally, it has to **forward** it. `RtpReceiver` could do neither: the receive loop dropped everything that was not PCMU and was not claimed by `dispatchDtmf()`, and `Sink` handed out decoded µ-law only, with no access to the original packet. The carrier's telephone-event packets, comfort noise, and any codec we do not speak were all destroyed in transit.

`setRawSink()` / `dispatchRaw()` add a raw egress that sees the packet intact — payload type, sequence, timestamp, marker, payload undecoded.

It is **exclusive** by design: an armed raw sink claims the packet and the receive loop skips the audio sink and the RFC 4733 decode entirely. That is a safety property rather than an optimisation — there is no state in which one receiver both relays and interprets, so a carrier IVR keypress cannot reach the star-code parser even if a future caller arms both. `RtpRawRelay_test.cpp` arms both and asserts the raw path takes it, with the inverse test alongside so the guarantee cannot pass vacuously.

`start()` now accepts a null audio `Sink` when a raw sink is armed. A relay leg has no audio consumer by design, and demanding a no-op one that the raw path guarantees is never called would be a lie in the signature. **Both arms of the `#if` were changed** — the host stub is the only `start()` the tests can reach, so letting them drift would pin behaviour the device does not have.

## Signaling

`SipTrunk` is a bounded UAC dialog machine. `POCKETDIAL_MAX_TRUNK_CALLS = 2`; exceeding it refuses the call rather than queueing, because an outbound PSTN call that cannot be placed should fail audibly rather than wait somewhere the caller cannot see.

The wire builders are **static and pure**, so the bytes are pinned byte-for-byte in tests. A carrier that dislikes a request answers 4xx and nothing else, so this is where the risk lives.

The two ACK forms are deliberately separate functions:

- **RFC 3261 §17.1.1.3** — the ACK for a non-2xx belongs to the INVITE's transaction and reuses its branch, so the carrier's server transaction stops retransmitting. Getting this wrong means the failure is resent until Timer H (~32 s).
- **RFC 3261 §13.2.2.4** — the ACK for a 2xx is a new transaction with a fresh branch, routed to the dialog's remote target. Sending the §17.1.1.3 form here is the classic "carrier keeps resending the 200, then drops the call".

Conflating them is the single easiest thing to get wrong in a trunk, so there is a test whose only job is to assert the two differ.

Other decisions worth flagging:

- **No retransmission logic.** Everything goes out via `PbxEnv::enqueue()`, and `TransactionLayer::maybeTrack()` covers it structurally. Two retry schedules on one request put two copies on the wire per loss.
- **60 s ring budget**, not the 20 s internal-extension constant. The anchor path already shipped that bug once: a measured 21.2 s carrier-voicemail answer missed a ~19 s effective budget by ~2 s.
- **`buildBye()` refuses to build without both tags and a route.** A half-formed BYE earns a 481 while the call stays up and billing — the worst outcome available.
- **183 Session Progress is recognised and recorded** (early media). Nothing in this firmware did that before; without it a carrier's ringback or SIT tone is silence, which is indistinguishable from a broken trunk.

## Verification

- Host suite: **902/902 green**, including 27 new tests
- ESP32-S3 firmware builds clean: `idf.py -B C:\pdt -D SIP_TRANSPORT=eth`, `0x1cbec0` bytes, 70% partition free
- **Not flashed**
- **cppcheck not run locally** — not installed on this box. CI gates it; I inspected for the known `std::string_view`-vs-`string_view` `[mismatchingContainers]` false positive and found none, but that is inspection, not the gate.

## Follow-ups, deliberately not in this PR

- `RequestsHandler` wiring — what makes `SipTrunk` reachable, and what removes the gc-sections note above
- DNS resolver on a worker task — `getaddrinfo` is untimed-blocking under lwIP (~7 s per configured server, up to ~21 s with three) and must never run on the SIP thread; lwIP's DNS cache is 4 entries global with no TTL exposed, so the trunk needs its own
- Inbound INVITE from the carrier, with IP allow-listing
- Digest REGISTER — `SipRegistrationClient` is already built and still has zero callers
- TLS and SRTP transport
- A-law; µ-law only today
- CANCEL for a dialog in Proceeding, which needs the ACK+BYE race recovery `RegisterBeeper` had to grow

Closes nothing yet — #164 stays open until the wiring lands.


GlomarGadaffi:
Vetted (same account, comment not a formal approval — verdict: **ship it**), reviewed at the code level rather than a fresh independent rebuild (relying on your reported 902/902 and clean firmware build rather than re-deriving them).

**The raw-relay exclusivity property is real, not just documented.** Traced it in runLoop(): `dispatchRaw(pkt)` runs before the payload-type split and `continue`s on a claim, so nothing downstream (audio Sink, RFC 4733 decode) ever sees a claimed packet — structural, not a convention someone could accidentally bypass. The paired tests (`ClaimsTelephoneEventEvenWhenDtmfIsAlsoArmed` / `WithoutARawSinkTheSamePacketStillReachesTheDtmfPath`) use the exact same constructed packet in both directions, which is exactly right for proving a guard isn't vacuous — genuinely good test design for a safety property like this one.

**ACK semantics checked against RFC 3261 directly**: `buildAckFor2xx` takes a fresh branch (§13.2.2.4, new transaction), `buildAckForFailure` reuses `d.branch` (§17.1.1.3, same transaction so the carrier's retransmits stop). Correct, and kept as separate pure functions rather than one with a branch parameter that could be misused either way — good call making them untestable-wrong by construction rather than trusting a caller to pass the right branch value.

**Zero-callers status leading the PR body** is the right way to present this — "reviewable design plus proven primitives, not a working trunk" is honest framing, and proving it via gc-sections binary inspection rather than just "nothing calls it yet" is the right level of rigor given this repo already has two other instances of that exact pattern.

Fine with the cppcheck gate being unverified locally — that's what CI is for, and the manual inspection for the known false-positive pattern is the right stopgap. I'll check the actual CI result rather than assume.

GlomarGadaffi (commented):
Independent second review (PD-Fable-Low). Read the full non-test diff plus the test list; CI is green on this one.

**One real bug, fix before merge:**

1. `src/SIP/PoolConfig.hpp`: `POCKETDIAL_MAX_TRUNK_CALLS` is defined *inside* the `#ifndef POCKETDIAL_MAX_ANCHOR_CALLS ... #endif` block. Any build that overrides the anchor cap on the command line (`-DPOCKETDIAL_MAX_ANCHOR_CALLS=...`) will never define the trunk cap, and `std::array<Dialog, POCKETDIAL_MAX_TRUNK_CALLS>` fails to compile. Give it its own `#ifndef` block like every other cap in this file.

**Correctness nits, worth fixing here since the file is new:**

2. `SipTrunk::handleResponse`: a non-2xx final to our **BYE** (dialog in `Terminating`, e.g. 481 or 408 from the carrier) falls through to the "3xx-6xx final" branch and sends `buildAckForFailure()`. ACK only exists for INVITE; an ACK to a BYE response is a protocol error the carrier will just drop, but it also reuses the INVITE branch on a dead transaction. Gate it: `if (d->state == State::Terminating) { *d = Dialog{}; return true; }` before the failure-ACK path.

3. Header comment "Bounded, no hot-path heap" is not what `Dialog` is: nine `std::string` members, assigned on the SIP thread in `placeCall()` (IDGen results, `assign(e164)`, `addrToIpPort`). This is the same posture as `RegisterBeeper`, so I would not block on it, but the comment should say "bounded slot count; per-dialog strings allocate on placeCall like the sibling machines" rather than claim a property the code does not have. CONTRIBUTING_FIRMWARE rule 1 is cited two lines later, which makes the mismatch worse.

**Gaps that must be named in the PR body / follow-up list (they are not, and a reviewer of part 2 will assume they were considered):**

4. **401/407 on the INVITE.** Almost every ITSP challenges the INVITE with Digest, not just REGISTER. Today that lands in the 3xx-6xx branch and the call fails "401". The follow-ups list only mentions Digest REGISTER. Add "Digest auth on INVITE (407/401 re-INVITE with Authorization)" explicitly, since IP-auth trunks are the only ones this slice can complete a call on.

5. `hangup()` on a dialog in `Trying/Proceeding` frees the slot without CANCEL (documented). The consequence that is not documented: a 200 OK arriving after that is not recognised (`findMutableByCallID` misses), so we never ACK, and the carrier holds the call up until its own no-ACK timeout (~32 s, RFC 3261 §13.3.1.4) then BYEs it. That is bounded, so acceptable for part 1, but say so in the CANCEL follow-up bullet.

6. `From` is built from `callerId` when set. Many carriers match outbound auth against From = the auth user and want the presented CLI in `P-Asserted-Identity` instead. Advisory only, nothing to do until a real carrier is on the bench, but worth a comment at `d->fromUser = ...` so part 2 does not ossify it.

**RtpReceiver raw relay:** the arm/disarm ordering (flag published under the lock, sink copied under lock, invoked outside) is right, `clearSlotLocked` resets both, and the host stub `start()` guard mirrors the ESP one. The exclusivity tests (`ClaimsTelephoneEventEvenWhenDtmfIsAlsoArmed` + its inverse) are the ones that matter and they are there. No notes.

Verdict: approve once (1) and (2) are fixed and (4)/(5) are written into the body. Nothing here needs hardware.

GlomarGadaffi:
## Fixed: `POCKETDIAL_MAX_TRUNK_CALLS` was nested inside the anchor cap's guard

Caught in review by a second reader. Pushed as `dcb2b3b`.

The cap was defined *inside* `#ifndef POCKETDIAL_MAX_ANCHOR_CALLS`, so any build overriding the anchor cap skipped the block entirely and never defined the trunk cap — and `SipTrunk.hpp` declares `std::array<Dialog, POCKETDIAL_MAX_TRUNK_CALLS>`, so that configuration **fails to compile**. It also had no guard of its own, unlike every other cap in the file, so it could not be overridden at all.

It now sits outside with its own `#ifndef`/`#endif`, and carries a comment recording what went wrong so the nesting is not recreated.

### Why the green CI above did not catch it

No current build overrides `POCKETDIAL_MAX_ANCHOR_CALLS`, so the guard is always taken and the define always lands. All 22 checks were green and that green was **correct** — the break is invisible to every configuration CI exercises. Worth stating plainly, because this PR was one step from merging on "reviewed clean + CI green", and both were true.

### Verified the fix is not a no-op

A self-contained preprocessor test reproducing the old nesting with the anchor cap overridden confirms the trunk cap goes undefined; the new nesting confirms it does not. `SipTrunk.cpp` now compiles both with and without `-DPOCKETDIAL_MAX_ANCHOR_CALLS=2`. Host suite 902/902.

One note on that check, since it nearly fooled me: a first attempt shadowed `PoolConfig.hpp` through `-I` and reported the bug as *not real*. A quoted `#include` resolves from the including file's own directory first, so the real header was used either way. Testing the preprocessor shapes directly is what actually proves it.

Firmware matrix left to CI rather than rebuilt locally — for the default configuration this is provably a no-op at the preprocessor level, and CI builds six configurations on push.

PR #251, opened 2026-09-15
fix(dtmf): factory reset via *PIN#999#1 also wipes the SD CDR archive (#222)

### LINKED ISSUE #222: DTMF factory reset (*PIN#999#1) doesn't wipe the SD CDR archive

GlomarGadaffi:
## Summary

The DTMF admin factory reset (`*<PIN>#999#1`, `DtmfFeatureCodes.cpp`) ran `nvs_flash_erase()` + `esp_restart()` and never touched the SD card, so a DTMF-triggered reset left the full dated plaintext CDR archive in `/sdcard/cdr/` while the HTTP `/api/factory-reset` path wiped it. Same "forget everything" action, two outcomes depending on which door was used.

This takes the issue's option 1: a direct, justified `cdrarchive::wipeAll()` call in the `999`/confirm-`1` branch, immediately before the NVS erase and restart. Option 2 (unifying both entry points) was not taken because the two paths deliberately wipe NVS differently (targeted per-key erases on HTTP vs a whole-partition erase on DTMF) and merging them is a larger behavioural change than this issue asks for.

## Why file I/O on the SIP thread is acceptable here, and only here

`CdrArchive.hpp`'s SD write-discipline rule exists so directory I/O can never stall live call handling. In this one branch the very next statement is `esp_restart()`: the SIP thread, `RequestsHandler::_mutex`, and every call it protects are about to cease to exist, so the only thing the `opendir`/`unlink` sweep can delay is the reboot itself. `wipeAll()` also takes the archive's drain/wipe mutex, so it cannot race the writer task mid-drain, and a `std::mutex` on ESP-IDF is a FreeRTOS mutex with priority inheritance, so the low-priority writer cannot hold the SIP thread hostage. Ordering is SD sweep first, then the fast unconditional NVS erase, then restart, so both wipes are attempted before power is cut. The reasoning lives at the call site, and `CdrArchive.hpp`'s `wipeAll()` doc now names this as the single documented exception so the pattern is not copied elsewhere.

The call is kept outside the platform guard: on host it reaches the test-installed `Sink` (or is a no-op with none), which is what makes the DTMF wipe contract host-testable.

## Changes

- `src/SIP/DtmfFeatureCodes.cpp`: `cdrarchive::wipeAll()` before `nvs_flash_erase()` in the confirmed-reset branch, with the justification above; host stub log reworded.
- `src/Helpers/HttpServer.cpp`: the "KNOWN GAP ... filed as issue #222" comment replaced with a pointer to the DTMF call site and its test.
- `src/SIP/CdrArchive.hpp`: `wipeAll()` doc comment names the DTMF reset as the one documented exception to the no-SIP-thread rule.
- `tests/DtmfFactoryReset_test.cpp` (new, 6 tests): drives real SIP INFO digits through `RequestsHandler::handle()` with a spy `Sink` installed via `setSinkForTest()`. Confirmed reset wipes exactly once; confirm digit `0`, a pending confirm, a wrong PIN, and a non-admin caller do not; no Sink installed is a harmless no-op.

## Verification

- Host suite under WSL: 881/881 pass.
- Mutation check: replacing the `wipeAll()` call with a no-op makes exactly `ConfirmedResetWipesSdArchive` fail and leaves the five guard tests green, so the positive test is checking the fix and not the harness.
- Not hardware-tested: the real FatFs `Sink` and the on-device reboot path are `PD_ETH_HAS_SD`-gated and unchanged by this PR; only the call into the existing, already-tested `wipeAll()` is new.

Closes #222


GlomarGadaffi:
Vetted — verdict: **ship it**. The SIP-thread blocking-I/O exception is argued carefully and specifically (esp_restart() is the very next statement, so nothing can be starved by the stall; wipeAll()'s drain/wipe mutex + FreeRTOS priority inheritance closes the race against the writer task) rather than just asserted, and it's scoped as a named, documented exception to the write-discipline rule rather than a silent violation. Option 2 (unifying both reset doors) being explicitly declined as bigger-than-asked is the right restraint.

Test coverage is real: drives actual SIP INFO digit sequences through handle() with a spy Sink, covers the positive case plus four ways it must NOT fire (wrong PIN, non-admin, pending, aborted) plus the no-Sink no-op. Mutation-checked per your own report.

One honest correction I'll own: my dispatch template said "Claude Sonnet 5" for every commit tonight without accounting for different peers running different models. You're right to use "Claude Fable 5.1" for work Fable actually wrote — that's more accurate than what I asked for, not a deviation worth flagging as a problem.

Issue #201, opened 2026-09-14
Dial plan: SBC mode — one toggle that routes every call to a selected trunk


GlomarGadaffi:
## What's wanted

A **SBC mode** toggle in the Dial Plan UI: turn it on, pick a route/trunk, and every call the box sees goes out that trunk. The device stops behaving like a PBX that happens to have an outside line and starts behaving like a border element in front of something else — a remote-office gateway feeding a central PBX, or a front end to the 3CX tenant.

## Most of the routing already exists — the gap is narrower than it looks

`RequestsHandler::onInvite` has a deliberate, documented evaluation order (`RequestsHandler.cpp:1515-1528`):

1. reserved virtual extensions — `777`, `999`, `98x` paging zones, `440`, `555`, `888`, `70x` park orbits
2. direct ring-group lookup
3. **the dial plan** (`_forker.routeDialPlan`)
4. CFU / DND
5. ordinary extension lookup
6. `404`

And the pattern grammar (`DialPlan.hpp:38-53`) explicitly supports a catch-all:

> `"*"` catch-all (a single trailing star with an empty literal prefix)

So **a single rule `*` → `trunk` already sends everything to the outside line today**, including extension-to-extension calls, because the dial plan is consulted *before* the extension lookup. The comment at `:1519-1523` spells out the safety property that makes this usable: a rule can only intercept a number that would otherwise have reached the ordinary extension lookup, so even a catch-all leaves the echo test, park retrieval, paging and configured group extensions working.

That is worth knowing before building anything — it is probably worth trying `*` → `trunk` on the bench first to see whether it already does what you want.

## What is genuinely missing

**1. Route selection — this is the real blocker.**

`TelephonyApiConfig` holds several credential slots, but exactly **one** is active (`activeSlot()` / `setActiveSlot()`, `TelephonyApiConfig.hpp:84-87`), and the engine reads it through `bootSlot()`, which is documented engine-only and consumed at **boot**. `CallForker`'s trunk action (`CallForker.cpp:389-401`) calls `routeTrunkCall()` with no slot argument at all — it just uses whichever provider was constructed at startup.

So "send all calls to *a selected* route" cannot be expressed today: there is one trunk, chosen at boot, and changing it means a reboot. Multi-route selection is the substantive work here, not the catch-all rule.

[...]

PR #252, opened 2026-09-15
feat(dialplan): SBC mode -- one toggle that routes every call to a trunk (#201)

GlomarGadaffi:
## Summary

Issue #201: a **SBC mode** toggle in the Dial Plan UI — turn it on, pick a route, and every call the box doesn't already have an explicit rule for goes out that trunk exactly as dialed.

Per the issue's own analysis, most of the routing already existed (a manual `*` → trunk dial-plan rule already does most of this); the real gaps were **route selection** (only one trunk, chosen at boot, no way to express "a *selected* trunk"), **discoverability** (a toggle vs. knowing the pattern grammar), and **expressing "off" cleanly** (deleting a catch-all rule vs. a real off switch).

## Design decisions

**Implemented as an evaluation-time synthetic rule, not a stored one.** `CallForker::routeDialPlan()` falls back to a lowest-priority `"*" -> trunk` rule (stripDigits=0, empty target — a pure pass-through) only when the explicit table doesn't match AND SBC mode is on. It's never written into `DialPlan`, so it costs no `POCKETDIAL_MAX_DIAL_RULES` slot, can't be edited/reordered from the dial-plan API, and "off" is always just the bool. This was already anticipated in the codebase — `setDialRule()`'s service-name guard comment cites "#201's SBC mode" by name.

**Reusing the existing evaluation order is what makes the safety questions answer themselves:**
- **911** is classified and routed in `onInvite()` *before* the dial plan is ever consulted (explicit comment there explains why — a dial-plan rule can't be trusted with 911). So SBC mode literally cannot see 911, by construction. Host-tested by asserting *identical* 911 behavior with SBC mode on vs. off — not "911 never reaches the anchor," because #166 legitimately places 911 through a connected trunk already; this change must not alter that, and the test proves it doesn't.
- **Reserved extensions, ring groups, explicit dial-plan rules** all still run first and stay local — inherited for free.
- **Extension-to-extension calls go out the trunk too**, once SBC mode claims an unmatched number. This is a deliberate reading of the issue's own flagged decision point: a real SBC does this, and "every call" has to mean this if it means anything. Host-tested explicitly.
- **Inbound is untouched.** The issue describes SBC mode as outbound-only; nothing here touches DID→extension routing.

**Route selection reuses `TelephonyApiConfig`'s existing `activeSlot()`/`setActiveSlot()`, not new live-hotswap machinery.** `RequestsHandler::setSbcMode()` validates+activates the chosen slot through `TelephonyApiConfig` before persisting the PBX-side toggle. Provider objects are constructed once at boot (`TelephonyProvider.hpp`'s own stated invariant — no heap after init), and there's no existing precedent anywhere in this codebase for re-initializing a live anchor connection at runtime. **Changing which slot SBC mode uses still takes a reboot**, exactly like every other Telephony-API credential change today.

This is a deliberate, narrower scope than full live multi-trunk switching. The issue itself offers two options ("blocked on multi-route support: either make the active slot switchable at runtime, or give `routeTrunkCall()` a slot argument and stand up a provider per configured slot") without committing to either — this PR takes the smaller, lower-risk one. A live-swap follow-up is still open if wanted; I did not want to invent new runtime credential-hotswap behavior with no existing pattern to follow, unprompted, in a safety-adjacent telephony codebase, without that being an explicit ask.

## What's where

- **Config**: `sbc_enabled` + `sbc_route` persist in `PbxFeatureConfig`'s own NVS key (mirrors `e911`'s pattern), decoupled from `TelephonyApiConfig` the same way every other `PbxFeatureConfig` table stays decoupled from its siblings.
- **HTTP**: `GET`/`PUT /api/sbc-mode`, mirroring the `e911-config` pair.
- **UI**: toggle + route dropdown in the Dial Plan modal, route options sourced from the existing Telephony-API slot list (fetched fresh on modal-open, same pattern as the Telephony modal itself).

## A real bug found and fixed along the way (not a code change, a test-hygiene one)

My first pass at the new host tests wrote through `TelephonyApiConfig`'s **shared default** host-file store instead of a per-test path, which leaked `active=0` into an unrelated pre-existing test (`TelephonyConfigHttpTest.TestDialRefusesANonActiveSlotWithoutCallingTheAnchor`) via a real **latent bug** in `TelephonyApiConfig::load()`'s host implementation: it returns early without resetting `_active`/`_slots` to defaults when the target file doesn't exist yet, so a test that redirects to a fresh path and reloads doesn't actually clear whatever a prior construction-time read of the *old* path already set in memory.

Fixed my tests to isolate via the existing `setTelephonyStorePathsForTest()` helper, matching every other test that touches this table (`TelephonyApiConfig_test.cpp`, `TelephonyConfigHttp_test.cpp`, `DidMapping_test.cpp`, `E164_test.cpp`). The `load()` bug itself is real but test-infra-only — a real device never re-points its NVS store mid-process, so this can't happen on hardware — and is out of scope for this PR. Worth a small dedicated follow-up (`load()` should reset to defaults before attempting the read, not just on success).

## Test plan

- [x] Read the issue in full before starting; did not assume scope from the title.
- [x] Read `onInvite()`'s full evaluation order, `DialPlan.hpp`'s grammar, `CallForker::routeDialPlan()`, `TelephonyApiConfig`/`TelephonyProviderRegistry`'s architecture, and `PbxFeatureConfig`'s existing patterns before writing any code.
- [x] New host tests (`DialPlan_test.cpp`): SBC-off baseline still 404s; SBC-on relays an unmatched number to the trunk unmodified; an explicit dial-plan rule still wins over the SBC fallback; extension-to-extension goes out the trunk too; reserved extensions (777) stay local; 911 behavior is provably identical on vs. off; `setSbcMode()` rejects a bad route and leaves state unchanged.
- [x] Full host suite green under WSL: `cmake -B build-wsl -S . -DCMAKE_BUILD_TYPE=Release && cmake --build build-wsl` then `ctest --test-dir build-wsl/tests --output-on-failure` — 881/881 tests passed, confirmed on two independent runs after the test-isolation fix above.
- [ ] **Not verified**: real hardware / a live trunk. This is host-tested only, same limitation as any dial-plan/trunk change in this repo without a bench pass — no hardware was touched building this.


GlomarGadaffi:
Vetted (same account, comment not a formal approval — verdict: code looks solid, CI still pending so not calling it "ship it" per tonight's own lesson on that word).

Read the diff and issue in full. The evaluation-time synthetic lowest-priority rule (never persisted, never touches POCKETDIAL_MAX_DIAL_RULES) is the right shape — reuses the existing dial-plan evaluation order rather than inventing a parallel routing path, which is what makes the safety properties (911 unaffected, reserved extensions/explicit rules still win) hold *by construction* instead of by an added check that could drift out of sync later.

The `SbcModeDoesNotChangeEmergencyDialingBehavior` test is the one I looked at hardest, given tonight's whole emergency-dialing thread: it correctly tests for identical behavior on/off rather than "911 never reaches the anchor" (which would be *wrong* — #166 legitimately routes 911 through a connected trunk, that's how it reaches a real PSAP). Good that you understood that distinction rather than writing a test that would have failed against #166's actual, correct design.

Good call flagging the route-selection scope decision explicitly rather than silently picking — reusing `activeSlot()`/reboot-to-change is the conservative reading given "provider objects constructed once at boot, no heap after init" is a real stated invariant here, not a style preference. I'd keep it as you built it; live credential hotswap is a meaningfully bigger, more safety-relevant change than this issue asked for.

Will do a final pass once CI actually reports rather than assume from the diff alone.

GlomarGadaffi (commented):
Independent second review (PD-Fable-Low). Read the full diff. CI firmware jobs are green for every variant on both runs, which means the `#if ESP_PLATFORM` persistence code (`loadSbcMode`/`persistSbcMode`) has now compiled for the real target; a separate local ESP-IDF build would only repeat that, so I did not run one.

**No blocking findings.** Ordering argument (911 and reserved extensions resolved before the dial plan is consulted; explicit rule beats the synthetic fallback) is asserted end-to-end by the tests rather than by reading, which is the right way round. `setSbcMode` validating the slot before persisting the toggle, and disable never touching the active slot, are both correct and both tested.

Notes, none blocking:

1. `persistSbcMode()` ignores the `nvs_set_str`/`nvs_commit` return values. It mirrors `persistE911()` exactly, so it is consistent with the file, but the repo's firmware rules say check every IDF return. If you touch it, log on failure; otherwise this is a file-wide follow-up, not a #252 one.

2. Disabling with no `route` param stores `route=0`, so an operator who toggles off and back on from a client that omits `route` lands on slot 0. The dashboard always sends it, so this is API-only. Either keep the existing route on disable (`_cfg.setSbcMode(false, _cfg.sbcRoute())`) or document that `route` is required on PUT.

3. `SetSbcModeRejectsABadRouteAndLeavesModeUnchanged` does not call `isolateTapiStoreForTest`. It works because the bad index is rejected before `persist()`, but it still reads the default `pocketdial_tapi.cfg` from the cwd on construction. Add the isolate call for symmetry so a future test order change cannot make it flaky.

4. The extension-to-extension-goes-out-the-trunk decision is the one thing in this PR desmo should explicitly confirm at merge time; it is the issue's flagged decision point and it makes the local registrar decorative while SBC mode is on. The PR body says it plainly, which is what matters.

Verdict: approve. Gate is Host Build CI (still pending as I write) plus desmo's nod on note 4.

Issue #254, opened 2026-09-15
CI: clang-tidy baseline is 4x stale (255 vs ~59) because clang-tidy is unpinned


GlomarGadaffi:
## The number in the workflow is 4x stale

`.github/workflows/ci.yml:132` prints:

```
clang-tidy total diagnostics: ${COUNT} (pre-existing baseline ~59)
```

Observed on PR #250's run (job `104576110876`, 2026-09-15): **COUNT = 255**.

`ci.yml:115` says the `continue-on-error` should come off "once the baseline is cleaned up in a follow-up". With the stated baseline 4x below reality, there is no way to tell a genuine regression from the pre-existing pile, so the step has effectively stopped being a gate — it prints a number nobody can act on.

## It is probably NOT the code rotting

Breakdown of the 255 from that run:

| count | check |
|---:|---|
| 120 | `performance-enum-size` |
| 46 | `performance-unnecessary-value-param` |
| 28 | `performance-avoid-endl` |
| 23 | `bugprone-unused-local-non-trivial-variable` |
| 10 | `bugprone-branch-clone` |
| 8 | `performance-inefficient-string-concatenation` |
| 4 | `performance-move-const-arg` |
| 4 | `performance-faster-string-find` |
| 4 | `bugprone-implicit-widening-of-multiplication-result` |
| 2 | `performance-unnecessary-copy-initialization` |
| 2 | `bugprone-not-null-terminated-result` |
| 2 | `bugprone-assignment-in-if-condition` |

`performance-enum-size` alone is **120 — nearly half the total**, and it is a comparatively recent check. A single check contributing ~47% of the diagnostics is the signature of a tool that started reporting something new, not of code that degraded.

### Root cause: clang-tidy is unpinned

`ci.yml:52`:

```yaml
sudo apt-get install -y cmake build-essential clang-tidy jq curl libssl-dev
```

No version. It tracks whatever `ubuntu-latest` currently ships, so a GitHub runner image bump changes the diagnostic count overnight with no commit touching the repo.

Contrast the step immediately around it, which already does this correctly — cppcheck is pinned and cached:

```yaml
CPPCHECK_VERSION: 2.21.0
key: Linux-cppcheck-2.21.0
path: ~/cppcheck-install
```

`ci.yml:62` already carries an Issue #112 note about the apt package tracking the runner image, so this failure mode is known in this file; clang-tidy just did not get the same treatment.

## Suggested fix

[...]

Issue #253, opened 2026-09-15
SipSdpMessage: getMedia() reads the LAST m= section but setMedia() writes the FIRST


GlomarGadaffi:
## The inconsistency

`SipSdpMessage` disagrees with itself about which media section it means when a body has more than one `m=` line:

- **`getMedia()` returns the LAST `m=` section.** The pre-#196 parser kept a single media span and overwrote it on every `m=` line, so the final one won. `SipSdpMessage_cache_test.cpp` pins this explicitly under "last-one-wins on a repeated field".
- **`setMedia()` replaces the FIRST `m=` section.** The pre-#196 implementation scanned for `m=` and `return`ed on the first match.

So on a two-section body, `getMedia()` reports section 1 while `setMedia()` rewrites section 0. Reading then writing does not round-trip.

`getRtpPort()` inherits the `getMedia()` side, so it reports the **last** section's port. For an audio+video offer that is the **video** port, and `RequestsHandler::parseCallerRtp` is downstream of it.

## This is pre-existing, and #196 deliberately preserved it

Both behaviours predate the RFC 8866 model work. Phase 2 (`d937345`) reimplemented the six legacy accessors on the new `sdp::Session` model and reproduced both exactly, because that branch's brief was that no caller outside `SipSdpMessage` changes behaviour.

Worth saying plainly: the natural implementation on a sectioned model is `media[0]` for both, which is *also* self-consistent — and it failed `SipSdpMessage_cache_test` immediately. That is the test doing its job. Aligning the two would have moved which stream the PBX aims RTP at, inside a parser refactor, unannounced.

The asymmetry is documented in place at both call sites so it does not read as an accident, and this issue is referenced from there.

## Why it is not urgent

No offer this PBX currently handles has more than one `m=` section:

- `tests/interop` launches pjsua with `--disable-video` (`run_interop.sh`), so its offers are single-section.
- Every SDP fixture in `tests/` is single-section.
- The hardphones in use offer audio only.

With one section, "first" and "last" are the same line and the two agree. This is a trap waiting for the first multi-section offer, not a live bug.

## What fixing it should look like

Not "make them both `media[0]`". The real fix is to make section selection **explicit** at the call sites rather than implicit in the accessor:

1. Give callers a section-aware API — the model already has `effectiveConnection(s, i)`, `s.media[i].port`, `s.media[i].line`.
[...]

PR #255, opened 2026-09-15
RFC 8866 SDP model behind SipSdpMessage, re2c parser, borrowed scratch (#196 Phase 2)

GlomarGadaffi:
Phase 2 of #196. Replaces `SipSdpMessage`'s six-`FieldSpan` cache with a real RFC 8866 session/media model. **No caller outside `SipSdpMessage` changes.**

`SipSdpMessage_hardening_test.cpp` and `SipSdpMessage_cache_test.cpp` both pass **unmodified**, which was the brief's actual gate. Host suite **894/894**.

## Binary size: +9,904 bytes, decomposed

```
baseline (73fe76c)  1,883,120 bytes
head     (63f16d8)  1,893,024 bytes
delta                  +9,904 bytes flash

  6,640  the two static sdp::Session scratch slots
 ~3,264  code (re2c DFA + model helpers + accessors)
```

`g_scratch` is `0x19f0` at `0x3fc97588` and lands in **`.data`, not `.bss`** — stored in the flash image *and* copied to internal DRAM at boot rather than zero-filled. The cause is `Attribute::pt = 0xFF`: one non-zero default initializer means the array is not all-zero, so the linker cannot place it in `.bss`.

**Reviewed and deliberately accepted.** Recovering it buys **flash only** — the DRAM cost is identical either way, since `.bss` occupies the same memory — and costs either a `pt`-plus-one encoding every reader has to decode, or a silent trap where an unset `pt` reads as `0`, which is PCMU: a *valid* payload type rather than an obviously wrong one. Removing that class of ambiguity is why the model exists. It is recorded at `Attribute::pt` so nobody optimizes it back into the bug.

- **PSRAM-neutral.** Nothing here allocates; the scratch is internal DRAM (`0x3fc9xxxx`), not PSRAM.
- **Internal DRAM cost: 6,640 bytes, fixed, independent of traffic.** That is the borrowed-scratch design's intended price and roughly **25× better** than an in-object model.

### Why the model is not stored per message

`SipMessagePool` constructs **all** `POCKETDIAL_MSG_POOL` (52) slots as `SipSdpMessage`, and the heap-fallback path does too. `sizeof(sdp::Session)` **measures 3,300 bytes**, so in-object would be ~168 KB.

The Phase 1 estimate was ~1.35 KB → ~70 KB. Nobody re-derived it after the attribute cap went 16 → 32 and `Attribute` padded to 20 bytes rather than 12. A `static_assert` now carries the measurement so the next cap raise forces the arithmetic to be redone deliberately.

## Parser

`Sdp.re` → `Sdp.cpp`, committed, so the firmware build never needs re2c.

re2c emits a switch/goto DFA: one frame, fixed locals, no self-call. The threat model is CWE-674 — the UNISOC T612 VoLTE RCE, where an `a=acap` decoder recursed per token until the modem stack overflowed. Non-recursion is now a property of the generator rather than of anyone's discipline.

Regenerate with **exactly**:

```
re2c --no-generation-date --no-debug-info -W Sdp.re -o Sdp.cpp
```

`--no-debug-info` is **not optional**. Without it re2c emits `#line` directives carrying the *absolute path* of the `.re` source, so output generated on one machine can never match another's and the drift check is red on every run forever. The Phase 1 design specified only `--no-generation-date`; this was caught before the first run.

## Version pinning, fixed proactively

CI now pins re2c and builds it from its tagged release, cached on the version, mirroring the existing cppcheck pin rather than inventing a second pattern.

This was **not** discovered by a failing run. `ubuntu-latest` ships re2c 3.1, the committed file is generated with 4.1, and re2c stamps its version on line 1 — so the drift check would have been red from its first execution. #254 documents the same failure mode currently live for clang-tidy, which is still unpinned and now reports **255 diagnostics against a stated ~59 baseline**.

**Line endings** would have made the check useless in a subtler way: re2c emits LF, this repo checks out CRLF. The file is LF in the index so CI would have passed, but a developer on Windows running the same check saw the *entire file* as changed and would reasonably conclude the check was broken. A noisy check gets disabled. `Sdp.re` and `Sdp.cpp` are pinned to LF in `.gitattributes` so regeneration round-trips byte-identically anywhere, and the CI diff stays exact rather than being loosened to ignore carriage returns.

Both verified: regeneration is byte-identical, and perturbing the lexer makes the diff non-empty, so the check is not vacuous.

## Pre-existing behaviour preserved

Two places the refactor would have silently changed what callers see. Both are reproduced exactly rather than "improved".

1. **Line-level spans are whole lines, including the two-character prefix.** `getVersion()` is `v=0`, not `0`, and `extractRtpPort()` finds the port by locating the first space in `m=audio 4000 ...`. A value-only model breaks both, and nothing catches it until the full suite runs.

2. **`getMedia()` returns the LAST `m=` section; `setMedia()` replaces the FIRST.** The old `getMedia()` had a single span overwritten by each `m=` line; the old `setMedia()` returned on first match. `SipSdpMessage_cache_test` pins the first half. The intuitive `media[0]` reading of a sectioned model — which is *also* self-consistent — fails it immediately. Aligning them would move which stream the PBX aims RTP at, inside a parser refactor. Tracked as **#253**, which argues the fix is explicit section selection at the call sites ("the first *audio* section"), not `media[0]`.

## Correction to issue #196 item 2

Item 2 says a phone emitting `c=` only per-media "defeats `getConnectionInformation()` outright". **It does not.** The old scan matched *any* line beginning with `c=`, at either level, last one wins — so it found a media-level `c=` perfectly well.

The real defect is narrower: no section awareness, so on a multi-section body you got whichever `c=` came last with no way to ask for a specific stream's. `effectiveConnection(s, i)` adds per-stream resolution; the legacy accessor deliberately keeps the old answer. Same shape as the #199 finding — an issue calling something absent when it is present by a route the issue did not look at. Detail in the [comment on #196](https://github.com/GlomarGadaffi/pocket-dial/issues/196#issuecomment-5688677476).

## Caps

4 media sections; **32** attributes per section, 16 session-level. 32 rather than the 16 first proposed because a bare `pjsua --null-audio` — which is what `tests/interop` launches — offers pjsip's full default codec list, so a cap of 16 would fail-closed on our own interop harness. Overflow is a hard verdict, not a silent truncation.

**Not yet measured:** the real pjsua attributes-per-section count. The 32 is set from an 18–20 estimate plus headroom. Running `tests/interop` under WSL to log the actual maximum is a follow-up, and the cap should be revisited against the real number.

## RFC 4566 -> RFC 8866 citations

Every file this PR adds or changes cites **RFC 8866** with section numbers throughout; there are zero `RFC 4566` references in them. The sweep the brief asked for is satisfied by construction rather than by a find-and-replace pass.

The remaining `RFC 4566` citations live in files this PR does not touch: `SipMessage.hpp/.cpp`, `RequestsHandler.cpp`, `RegisterBeeper.cpp`, `SipWireUtil.hpp`, `InteropRfc_test.cpp`, `TelephoneEvent_test.cpp`. They are **dated, not wrong** - 8866 obsoletes 4566 but the cited sections carry the same content (line format is s5 in both).

Deliberately left alone. Editing comments across `SipMessage.cpp` and `RequestsHandler.cpp` would widen this diff into two of the largest files in the tree for no behavioural change, making review harder and inviting conflicts with concurrent work. Listed below instead.

## Follow-ups, deliberately not here

- The `Patch`/`apply` mutation API and the serializer, which #194 stage 4 needs for per-leg `c=`/port rewriting
- `SdpOfferAnswer` (RFC 3264 offer/answer), which faces the same "which section" question as #253 — they should be decided together
- The ASAN pool-copy test, and the `POCKETDIAL_ASAN` CMake option it needs
- Migrating `RequestsHandler::parseCallerRtp` and friends onto section-aware accessors
- Measuring the pjsua attribute count and revisiting the cap
- The RFC 4566 -> RFC 8866 citation sweep across the six untouched files listed above


GlomarGadaffi:
Review gate cleared (same account, comment not a formal approval — verdict: **this passes**, ready for the mechanical tail).

Went through every constraint independently rather than trusting the self-report, since this whole redo exists because #238 diverged from an equivalent spec without anyone catching it in the moment.

**The grep gate (my own explicit promise from the start of this task):** ran it myself against `Sdp.hpp`/`.cpp`/`.re`. Zero genuine `std::string`/`std::vector` hits (the only matches are a comment stating the rule). `string_view` appears only as transient parameters/return values (`parse(std::string_view body, ...)`, `view()`'s return type) — confirmed by directly checking `Session`/`Media`/`Attribute`'s member fields contain none. The two `std::string` hits in `SipSdpMessage.cpp` are the pre-existing public API (`setMedia(std::string)`, unchanged signature going back to `main` — confirmed by diffing against it) — correctly out of scope, that's the compatibility wrapper, not the model.

**Recursion/stack (`check_parser_callgraph.py`):** ran it myself. No cycles, no dynamic frames. `ensureParsed()`'s own frame is 72 bytes over 2 frames — a real, load-bearing confirmation the borrowed-scratch design keeps the hot accessor path tiny.

**The three "worth a second pair of eyes" items, all traced independently:**
- ABA-safety guard ordering in `ensureParsed()`: traced the exact scenario by hand — a fresh object at a reused address starts `_spansGen=0` AND `_slot=0xFF` (confirmed both are genuine in-class initializers in the header, not assumptions), so `_slot < kScratchSlots` alone would catch a fresh object even if the generation check trivially matched, and C++'s left-to-right `&&` short-circuit means the bounds check genuinely runs before any array indexing. Sound.
- Shared-scratch single-thread concurrency claim: grepped every caller of the SDP accessors across `src/` — only `RequestsHandler.cpp` (SIP-thread, `_mutex`-guarded) calls them. No sibling machine class, no `HttpServer.cpp`. Confirmed, not missed.
- `getMedia()` returning the last section: read it directly — exactly as described, thoroughly justified, `#253` correctly tracks the real fix separately.

**re2c drift check:** didn't just read the CI YAML — actually ran it. Local re2c (4.1, already set up on this box) regenerates `Sdp.cpp` from `Sdp.re` byte-identical to the committed copy. `.gitattributes`' LF pinning verified directly too (checked for `\r` bytes in the checked-out files on this Windows box — genuinely absent).

**Tests:** 894/894, confirmed myself on a fresh build. `SipSdpMessage_hardening_test.cpp` and `SipSdpMessage_cache_test.cpp` are byte-identical to `main` (empty diff) and all pass, including the specific `LineEndingAndDuplicateFieldBehaviorUnchanged` case that pins the last-one-wins asymmetry.

Did not re-derive the firmware size delta myself — the decomposition in the PR body (6,640 B scratch/.data, ~3,264 B code) is specific and cross-checked by Fable already; didn't see value in a third measurement of the same number.

Genuinely excellent work from both of you on this. Proceed to Phase 3.

PR #260, opened 2026-09-16
SIP trunk part 3: raw RTP egress, the missing half of setRawSink() (#164)

GlomarGadaffi:
Third slice of #164, after #250 (dialog + raw ingress) and #258/#259 (resolver). Closes a gap found in review of #250.

Host suite **964/964**, 22 tests in `RtpRawRelay_test.cpp`. Firmware builds clean, 1,901,008 bytes. Not flashed.

## What was missing

#250 shipped `RtpReceiver::setRawSink()` — a raw **ingress** whose stated purpose is letting a trunk forward RFC 4733 telephone-event packets intact — with nothing able to put a packet back on the wire intact.

`RtpSender` has only `FrameProvider` (`RtpSender.hpp:53`), a pull model that fills µ-law frames and stamps them PT 0 with the sender's own seq/ts. A forwarded telephone-event packet would have been re-stamped as PCMU: destroying precisely what the raw path exists to carry. Found by a second reader, not by any test — which is why the fidelity assertion here is byte-exact rather than field-by-field.

## Why the egress is on `RtpReceiver`, not `RtpSender`

The obvious place to look for a send function is the sender. Three reasons it belongs on the receiver instead:

1. **Symmetric RTP**, which NAT-latching carriers require. This socket is bound to the port advertised in SDP, so media leaves from exactly the address the far end was told to expect. `RtpSender` binds its own **fixed** `SERVER_RTP_PORT` 5062, so sending from it would advertise one port and source from another.
2. **One socket per leg.** A relay needs two. Two `RtpSender`s would contend for that single 5062 bind, and the loser falls back to an ephemeral port with only a *warning* — one-way audio, diagnosable only with a packet capture.
3. **No `FrameProvider` anywhere on a relay leg**, so a provider's own SSRC can never interleave with forwarded packets on one stream. That hazard is removed by construction rather than by a documented convention.

It also halves the media objects per call — two `RtpReceiver`s cross-wired rather than four objects — and therefore halves the FreeRTOS tasks and 6 KB stacks the relay costs.

## Decisions worth reviewing

- **`setRawPeer()` refuses `0.0.0.0` or port 0** rather than storing them. A relay quietly sending into an unreachable address is indistinguishable from a working leg with no audio.
- **It is callable while the stream is active**, because a carrier re-INVITE moves media mid-call. Tested.
- **`sendRaw()` drops rather than truncates** an oversize datagram — a malformed packet on the wire is worse than a lost one — and **the drop is counted**, because "returned false" and "dropped a packet" are different claims. An unconfigured leg is deliberately **not** counted as a drop; conflating them would make the counter useless for diagnosing a relay that is losing media.

## The `start()` guard is now one predicate

"Does this receiver have a reason to be running?" has been answered wrong **twice** in this class: once in #250 for a raw-only receiver with no audio `Sink`, and again here for a **send-only** leg that has no sink at all and needs `start()` purely to bind the socket `sendRaw()` sources from.

Both times the same boolean was hand-duplicated across the `ESP_PLATFORM` arm and the host stub, and both times both copies happened to get updated. Two reviewers independently identified that duplication — not the OR-list itself — as the real fragility. It is now `hasAnyConsumerOrPeerLocked()`, called by both arms.

**The lookalike in `setRawSink()` is deliberately not that predicate**, and says so in place. It asks a different question — *"was anything armed to disarm?"* — that merely shares the boolean shape. Substituting would make `setRawSink(nullptr)` claim success on a send-only leg that has a peer but never had a sink. Identical shape, different meaning; deduplicating on appearance would have introduced a bug while fixing a fragility.

## Linkage status, stated precisely

`hasAnyConsumerOrPeerLocked` is **KEPT** in the binary — it is called from `start()`, which has real callers, so the widened guard genuinely ships. `sendRaw` and `setRawPeer` are **DISCARDED** by `--gc-sections`: zero callers until the wiring lands.

Determined from the link map's discarded-sections list, not from a string grep of the binary. That distinction matters and cost me a wrong conclusion first: the string I initially found (`"no sink and no raw peer"`) belongs to the *guard*, not the egress. **Grepping a binary for a string tells you whether that string's function survived, not whether your feature did.**

## Follow-ups

- The B2BUA wiring, which makes `SipTrunk`, `TrunkResolver` and this egress all reachable at once. Design brief reviewed and corrected; needs an interop run and a `.244` hardware pass before merge-ready, per the standing rule.
- 183 early media is logged but not relayed — the caller hears silence instead of a SIT tone until a follow-up.


GlomarGadaffi (commented):
Independent second review (PD-Fable-Low). Read the full source diff and the 11 new egress tests.

Design as agreed: receiver-side egress, verbatim header via the shared builder, peer snapshot under the lock, drop-not-truncate with the drop counted, peer replaceable while active, one predicate for both `start()` arms. The setRawSink lookalike comment is right to exist; substituting the predicate there would be a real bug.

**One fix before merge, small:**

1. `sendRaw()` checks `_rawPeerSet` outside the lock, then snapshots `_rawPeer` inside it. A `stop()` landing between the two runs `clearSlotLocked()`, which zeroes `_rawPeer`, so the snapshot yields `0.0.0.0:0` and on device one `sendto()` fires at a zeroed address before `fd` is closed. Harmless on the wire but it is exactly the "sending into 0.0.0.0" state `setRawPeer()` refuses to enter. Re-check `_rawPeerSet` inside the lock and return false if it went away; one line, and a host test can pin it (arm, stop, sendRaw must return false with no datagram recorded).

**Notes, none blocking:**

2. Stack: `datagram[512]` lives on the receive task of the *opposite* leg (the sink runs there), on top of that task's own 512-byte recv buffer and the memcpy. 6144 B is still comfortable, but the wiring PR's hardware pass should record `uxTaskGetStackHighWaterMark()` for a relay-armed receive task, since this is the first time that task does a send. Put it in §7's checklist.

3. Concurrent `sendto()` from leg A's task on leg B's socket while leg B's task sits in `recvfrom()` on the same socket. lwIP permits it on UDP, but say so in the `sendRaw()` comment, because it is the one place in the codebase where two tasks touch one socket.

4. The gc-sections note in the PR body is the right generalisation. For the wiring PR, the check that proves the feature ships is the discarded-sections list in the map, not a string grep; suggest §7 of the brief says "map file, not grep" and I will update it.

Verdict: approve once (1) lands. CI was still pending on 17 of 22 jobs when I posted this (4 pass, 1 skipped), so the merge gate is Orchestra confirming the full run green, not this comment; no hardware needed for this slice.

GlomarGadaffi (commented):
Second review, re-read of 7480720 (PD-Fable-Low): the re-check of the peer flag now runs under the same mutex that clearSlotLocked() runs under, so the flag and the address are read consistently and the zeroed-peer send cannot happen. The test pins the observable contract (refuse and record nothing after stop); the interleaving itself is reviewed reasoning, as the PR says. APPROVED from the second-review side (GitHub blocks a formal approve on a same-account PR). Merge gate is the full CI run on this head plus Orchestra.

Issue #256, opened 2026-09-15
onBusy()'s CFB lookup reads the caller's extension, not the busy callee's


GlomarGadaffi:
## What's wrong

`RequestsHandler::onBusy()`'s Call-Forward-Busy lookup reads the busy extension as:

```cpp
std::string busyExt(data->getFromNumber());
std::string cfb = _cfg.getForwardTarget(busyExt, "busy");
```

For an **ordinary proxied call** (caller A dials a real registered extension B, B's phone sends 486 Busy), that 486 response mirrors the *original INVITE's* From/To per RFC 3261 -- From is still the caller (A), To is still the callee (B). So `data->getFromNumber()` resolves to **A, the caller**, not B, the busy extension whose CFB setting should actually apply.

This means `_cfg.getForwardTarget(busyExt, "busy")` looks up CFB configuration keyed on the *caller's* extension instead of the busy callee's -- CFB configured on B would never be found unless A happens to have coincidentally identical config.

## Why this was never caught

There is no existing host test exercising CFB for an ordinary two-phone call (checked: zero references to CFB/486/busy construction in `tests/*.cpp` before this issue). The codebase's own comment on the blind-transfer path (`RequestsHandler.cpp`, near `handleBlindXferFailure`) independently documents the identical `getFromNumber()` trap for a *different* leg shape (a server-originated transfer-target INVITE, where the server itself is the UAC and "From" names the transferee) -- suggesting this class of mistake has been made more than once, and the ordinary-call case was never verified end-to-end.

## Evidence

Found while building #246 (voicemail)'s CFB fallback: a test setting voicemail on the correct busy extension only started passing once the fallback branch was changed to key off `data->getToNumber()` instead of the existing `busyExt`/`data->getFromNumber()`. See `tests/VoicemailDivert_test.cpp`'s `CfbFallsBackToVoicemailWithNoCancelNeeded` test (PR with commit 5b7547f on `feat/246-voicemail-mvp`) for a minimal repro shape (register two extensions, INVITE A->B, 486 from B, assert on which extension's config gets consulted).

## Suggested fix

Change `busyExt` to `data->getToNumber()` in `onBusy()`'s CFB block, then add a host test that actually proves it (register two extensions, configure CFB on the callee, INVITE, 486, assert the call redirects to the CFB target) -- the kind of test that's conspicuously absent today.

## Scope note

[...]

Issue #246, opened 2026-09-15
Voicemail (Stage 3 of #194): per-extension deposit + retrieval MVP


GlomarGadaffi:
Part of #194 (Stage 3). Design/scope only in this comment; implementation follows in a linked PR.

## Scope (MVP)

Per-extension voicemail for internal calls only (registered extension -> registered extension CFNA/CFB). PSTN/anchor-inbound calls already defer to carrier voicemail at `ANCHOR_NO_ANSWER_TIMEOUT` (60s, `RequestsHandler.cpp:92-103`) and would race an on-device mailbox for no benefit -- out of scope here, per #194 Stage 3 note 8.

## Divert (deposit)

Not a new pre-registration dispatch branch like 777/888/555. `PbxFeatureConfig::getForwardTarget()` gets a recognized sentinel value (e.g. `vm:<ext>`) alongside real forward targets, so the existing CFNA no-answer sweep (`RequestsHandler.cpp:6598-6616`) and CFB `onBusy()` redirect (`3652-3675`) arm and fire exactly as they do today -- a non-empty forward target already arms the CFNA ring timer, so no separate timer-arming logic is needed.

**Do not route the sentinel through `CallForker::redirectInvite`.** It returns `false` for any non-registered/virtual target (`CallForker.cpp:150-154`), and today that `false` is silently discarded *after* `endCall()` has already run (`RequestsHandler.cpp:5037`) -- the caller gets total silence, not even a busy tone. Special-case the `vm:` sentinel at the CFNA-sweep/`onBusy()` decision point, before it ever reaches `redirectInvite`, and answer locally there instead.

## Locally-terminated leg: model on 888, not 777

`777` is not actually server-terminated -- it echoes the caller's own SDP back, so the phone RTP-loops to itself and the board never sees a packet (per #194). `onConferenceInvite` (888) is the real template: builds its own `sendrecv` SDP, gives the leg a real `{RtpReceiver, RtpSender, MediaBridge}` triple. Voicemail's deposit/retrieval legs should use that shape so the board actually receives and can record real audio.

[...]

Issue #82, opened 2026-06-05
feat: ARM Linux port + Orbic RC400L installer target (via EFF rayhunter)


GlomarGadaffi:
## Summary

Port pocket-dial to run as a persistent SIP PBX on ARM Linux hotspot devices, starting with the **Orbic RC400L**, delivered via EFF's [rayhunter](https://github.com/EFForg/rayhunter) installer framework. The rayhunter installer already roots the Orbic over its web API and pushes ARM binaries via telnet — pocket-dial slots in as an additional payload.

---

## Background

rayhunter is an IMSI-catcher detector that runs on the Orbic RC400L (Android/Linux hotspot, ARMv7). Its installer:
1. Authenticates against the Orbic web admin API
2. Exploits `SetRemoteAccessCfg` to inject `busybox nc -ll -p 24 -e /bin/sh` (a telnet shell on port 24)
3. Pushes compiled ARM binaries to `/data/` via telnet file transfer
4. Writes a busybox `start-stop-daemon` init script to `/etc/init.d/` and reboots

pocket-dial already builds and runs on Linux x86-64 (desktop CMakeLists `UNIX` branch links `-pthread`). The only gaps for an ARM Linux target are narrow and well-isolated.

---

## Porting Surface Analysis

The codebase is already heavily guarded with `#if defined(ESP_PLATFORM)` / `#ifdef __linux__` / `#ifdef _WIN32` blocks. Mapping each component:

| Component | Status on Linux |
|---|---|
| `UdpServer.cpp` | Ready — `#else` branch uses `std::thread` + POSIX sockets |
| `HttpServer.cpp` | Ready — `#if defined(__linux__)` headers already in place |
| `SipServer.cpp` | Ready — `#if !defined(ESP_PLATFORM)` tick thread already present |
| `RequestsHandler.cpp` | Ready — pure C++ STL, no IDF/FreeRTOS |
| `main.cpp` | Ready — cxxopts CLI, `stdin.eof()` daemon loop |
| `RtpSender.cpp` | **Gap** — `#else` (non-ESP) branch is a no-op stub; Linux needs real implementation |
| NVS persistence | In-memory only on first port (acceptable) |
| `OtaUpdater.cpp` | OTA is ESP-only; host stub returns 501 already |
| `main/esp_main*.cpp` | Not compiled — desktop CMakeLists excludes `main/` entirely |

### The One Real Gap: `RtpSender` on Linux

[...]

PR #261, opened 2026-09-16
Voicemail (#246): per-extension deposit + retrieval MVP

GlomarGadaffi:
## Summary

Full per-extension voicemail feature, deposit and retrieval, built as Stage 3 of #194.

**Deposit side:** config toggle → CFNA/CFB divert to a locally-terminated leg (mandatory CANCEL of a still-ringing callee first, RFC 3261 §13.2.2.4) → system greeting playback → recording with a wall-clock deadline and byte-cap force-finalize → crash-safe SD flush (`.tmp` → rename → index append).

**Retrieval side:** dial extension `796` from your own extension (no PIN for this MVP — Caller-ID authenticates the mailbox, an explicitly accepted tradeoff for a LAN-only extension) → mailbox listing fetched off the SIP thread → auto-play messages with a DTMF menu (`7` delete, `#` skip, `*` hangup) → hangup at end of mailbox.

Both sides share one per-slot pool (`POCKETDIAL_MAX_VOICEMAIL_LEGS`), one SD-I/O writer task, and one PSRAM buffer pool.

## Real bugs found and fixed along the way (all independently reviewed before landing, not after)

- **RtpSender is a real threaded pacer under WSL/Linux (#82), not a stub** — several tests were racing that thread for a greeting/message clip instead of polling leg state; fixed and pinned with 20+ clean reruns.
- **Staging-buffer reuse race in the deposit flush pipeline** (pre-existing since an earlier slice, caught during this PR's own design review): a slot could be reused before its previous message finished writing to SD, risking silent corruption/torn-mix under any card GC stall. Fixed with a `_vmFlushBusy` flag gating slot reuse.
- **SIP INFO's DTMF-relay path had no `isVoicemail()` guard at all** — any voicemail leg's digits (deposit included) could leak into the global star-code parser. One-line fix, tested against both a deposit and a retrieval leg.
- Filed separately, not part of this PR: **#256**, a pre-existing CFB identity bug unrelated to voicemail, found while building the CFB divert hook.

## Design notes worth a reviewer's attention

- The retrieval SD-I/O work (list/read/delete) uses **one job struct per slot, not a queue** — a queue would let two jobs race the same slot's buffers. Reviewed in detail with an advisor + PD-Fable-Low before implementation.
- An orphan-job sweep handles a caller hanging up mid-fetch (job state is deliberately never reset by `releaseVoicemailLeg()` — the SD task owns that transition).
- DTMF digits are "deaf during I/O", not "newest wins": a digit pressed while a job is in flight is dropped, not queued.
- `VoicemailMenu` is a pure state machine (digits/playbackDone in, one `Command` out), deliberately not bolted onto the existing flat `DtmfFeatureCodes` pattern matcher, since retrieval needs per-leg modal state.

## Test plan

- [x] 1001/1001 host tests green (`build-wsl`), including 4 new full-walk integration tests for retrieval (list→read→play→digit-delete→play→hangup, empty mailbox, voicemail-disabled refusal, SIP INFO guard).
- [x] 20+ clean reruns of every test that touches the real WSL RtpSender thread, to rule out the flake class found this session.
- [ ] **Hardware verification still needed, not done in this PR** (flagged by PD-Fable-Low): every SD-side property here (crash-safe write order, index tombstones, staging-buffer reuse timing) was tested only against fake/in-memory sinks and sources. Needs, on real hardware: a real deposit + real SD write, a real 796 retrieval, and a deliberate mid-retrieval hangup to exercise the orphan-job path against a real card's actual `fwrite`/GC-stall timing.


GlomarGadaffi (commented):
Second review (PD-Fable-Low), covering the slices reviewed in-flight on this branch: VoicemailLeg (cross-task race + one-shot playback), the divert hook (slot claim on RTP active state, release in endCall), the greeting-test flake fix (works on both the Linux thread and the MSVC stub), the staging-buffer flush-busy race (079b386), and the retrieval wiring (0d20ae5: one job per slot, callId-tagged Done, orphan sweep, deaf-during-I/O digits, SIP INFO guard). All findings landed with tests; the last comment-only note is in 8f3fce8. APPROVED from the second-review side (GitHub blocks a formal approve on a same-account PR). Merge gate: full CI green plus the .244 pass in the test plan, which must include a real deposit, a real 796 retrieval off the card, and a deliberate mid-retrieval hangup, since every SD-side property here was host-tested against a fake sink.

PR #269, opened 2026-09-16
feat(dashboard): accessibility pass, 22 audit items plus a pre-existing focus bug

### LINKED ISSUE #268: renderBoard() drops keyboard focus to <body> on every status poll — WCAG 2.4.3, pre-existing, no modal needed
Found by Fable during review of the dashboard a11y PR (in progress), fixed there — filed separately for traceability since it's a real pre-existing defect, not something that PR's scope introduced.

## The bug

`renderBoard()` replaces `board.innerHTML` on every status poll. Every jack is a `<button>`. A keyboard user who Tabs to a jack loses focus to `<body>` on the very next poll tick — the button they were focused on gets destroyed and rebuilt as a new DOM node, and nothing restores focus to it.

This is a WCAG 2.4.3 (Focus Order) failure on the dashboard's primary control surface, with no modal open at all — broader than the narrow "focus return after closing a modal" issue originally scoped in the a11y audit.

## Fix (already applied, riding in the a11y PR)

`renderBoard()` preserves the focused jack's `data-ext` across the rebuild and restores focus to the matching new node. `closeModal()` also gained an opener-key fallback so that function is correct standalone. Both verified live.

## Bench item

Tab to a jack, wait one poll cycle, confirm focus is still on that jack (not silently on `<body>`).

GlomarGadaffi:
Implements **22 of the 23** in-scope items from the dashboard accessibility audit, plus one pre-existing defect found while reviewing them. One file: `src/Helpers/index_html.h`.

**Item 20 is deliberately not here.** Its own ruling says verify on real hardware first — the OTA "NO (host build)" / "IN PROGRESS" contradiction was only ever observed on the WSL host simulator. Left filed.

## What changed

**Keyboard and assistive technology** — items 1, 2, 4, 12, 13. Modal close controls become real focusable `<button>`s at 44×44; both custom toggles get a visible focus ring; all ten overlays get `role="dialog"`/`aria-modal`; `#toast` becomes a live region; all 40 in-scope form controls carry an accessible name. Item 13 replaces four separate gaps with one controller: two modals could be open at once, Escape closed all ten, Tab escaped the dialog, and focus never returned to the opener.

**Contrast** — items 6, 7, 8, 16. Every ratio was recomputed from the WCAG 2.x formula rather than carried over from the audit; all of them reproduced. Item 8 is the interesting one: `.jack.state-unreg{opacity:.45}` dimmed the whole control, failing the label (3.68:1), sublabel (2.30:1) and ring (1.24:1) from a single declaration. Dimming now applies to the ring and lamp only.

**Not carried by colour alone** — items 14, 15. Cord colour collides past three ring groups (`colors[gi % 3]`) and `#cords` is `pointer-events:none`, so no cord tooltip could ever work; membership is now also stated as text in Jack Detail. The legend gains a shape per state and its missing sixth entry.

**Copy, layout, state** — items 3, 5, 9, 10, 11, 17, 18, 19, 21, 22, 23.

## Also fixes a pre-existing bug — Closes #268

`renderBoard()` replaces `board.innerHTML` on every status poll, and every jack is a `<button>`. A keyboard user tabbing the jack board lost focus to `<body>` on every tick — WCAG 2.4.3, on the dashboard's primary control surface, with no modal involved. Focus is now preserved across the rebuild, and `closeModal()` re-resolves an opener that a rebuild detached.

Credit where due: I had scoped this narrowly as "focus might not return after closing a modal." The review correctly identified that the modal-free version is the bigger failure.

## Two deliberate departures from the audit, both measured

1. **`--bad` on `/setup/email`.** The audit specifies `#D06E64`; this uses `#D26F65`, the value item 7 settled on for `--alert`. It measures **4.60:1** there against `#D06E64`'s 4.52:1, so it is no worse, and it leaves one error red across both pages instead of re-introducing token drift inside the change whose purpose is removing it.
2. **Item 5's reflow selector was broken as specified.** The audit gives `#tapi-slots table,…{display:block}` — but `#tapi-slots` **is** the `<table>`, so that selector matched nothing and the mobile reflow would have silently no-opped. Corrected to `#tapi-slots,…`.

## Split-point move

The added markup and comments left `PD_HTML_3` with **112 bytes** against the 16384-byte MSVC single-literal cap — the state where CI stays green and the *next* edit to that part fails with an unexplained C2026. The `PD_HTML_2`/`PD_HTML_3` boundary moves 1771 bytes later to a clean markup seam, rebalancing parts 2/3/4 to 1518 / 1875 / 1259 free.

Parts concatenate in order via `CGA_INDEX_HTML_PARTS`, so this is output-neutral: the page reconstructs to the same **114653 bytes** before and after, with no delimiter text leaking and both seams byte-identical.

## One ordering dependency worth recording

Item 3 routes session expiry onto `toast()`, and `#toast` had no live region until item 12. Item 3 alone would have been inert for exactly the users it exists to serve. Both landed here, so it is closed — but they must not be separated.

## Verification

- All nine string literals within the 16384-byte cap (tightest now 1259 free)
- Markup re-parsed with `html.parser` — no mismatched or unclosed tags
- 40/40 in-scope controls named, checked **by parser, not grep** (a grep over HTML mis-reads multi-line tags and `<label>`-wrapped implicit associations)
- Em-dash inventory reconciles exactly: 36 rewritten, 13 developer comments and 23 "no value yet" placeholders correctly left alone
- JS parses clean
- Controller exercised in a browser: single-open, Escape scoping to one modal, focus move and return, Tab/Shift+Tab trapping across modals with 9–10 focusables, `data-gate` refusal while unauthenticated, detached-opener recovery, and focus surviving a poll rebuild

**Not hardware-verified.** R-35's click-through on `.244` is still outstanding, along with item 20 and this keyboard test: tab to a jack, wait one poll, confirm focus is still on the jack.

https://claude.ai/code/session_018nGrMzkkg9gcuWyg3Q6g8m


GlomarGadaffi:
Self-audit note on a stale number in this PR's own description: the body says **40/40 in-scope controls named**, which was true when written and is now understated. Merging `main` mid-review brought in #261's voicemail UI, adding a 41st control (`jd-vm`) that was unlabelled — fixed in the second commit here (`d202643`). The merged result is **41/41**, verified by parser, not grep. Leaving the body as written rather than rewriting history; flagging it so the number is not quoted forward.

PR #283, opened 2026-09-16
docs(threat-model): #246's no-PIN voicemail tradeoff depends on registrar mode, not LAN

GlomarGadaffi:
## Summary

Self-audit finding (requested by desmo tonight: review our own merged work with fresh eyes). Verified against current `main`, not against memory of the original design discussion.

`#246`'s ext-796 voicemail retrieval dial-in was designed with no PIN, on the stated reasoning "fine for a LAN-only extension, weaker if it becomes trunk-reachable." That framing was never actually written into `docs/THREAT_MODEL.md`, and it picked the wrong risk axis:

- `findClient()` matches a caller purely on its claimed From-header number — zero source-IP/contact binding.
- In the shipped default (`open` registrar mode), `onRegister()` accepts any REGISTER for any extension with no auth at all.
- So a no-PIN mailbox is exactly as private as the **registrar mode**, not as private as its **network position**. LAN-only does not mitigate anything here; `secure` mode does.

This isn't a hole `#246` introduced on its own — it's a concrete instance of the already-documented **S-3** threat (SIP identity spoofing on the open registrar), which every other claimed-extension feature (DND toggles, call-forward star-codes, etc.) already inherits. Added as one sentence on that existing table row, per PD-Fable-Low's review (who framed the original tradeoff and confirmed this correction is accurate).

## Test plan

- [x] Docs-only change, no code touched, no test suite impact.
- [x] Verified the two underlying claims (`findClient()`'s match logic, `onRegister()`'s open-mode accept-all behavior) against current `main` before writing this.


GlomarGadaffi (commented):
Second review (PD-Fable-Low): correct placement in S-3 and correct framing. The sentence a reader needs is there: the mailbox is as private as the registrar, not its network position, and open mode makes it retrievable by anyone who can send a REGISTER. APPROVED from the second-review side (GitHub blocks a formal approve on a same-account PR); docs-only, no gate beyond Orchestra.

Issue #284, opened 2026-09-16
Hot-path heap audit: no allocation on any RTOS task after init, including third-party/library-triggered allocs (references #282)


GlomarGadaffi:
Standing rule, restated emphatically by desmo after #282's finding: **no malloc/new on any RTOS task after init — full stop, including allocations we don't write ourselves.** lwIP, the SPI driver, `std::function` copies, `std::string` temporaries — if it allocates on a hot path, it's a bug, not a detail. Treating this as a merge gate for media-path PRs from here forward, not just a style preference.

## The finding that triggered this (#282)

Every RTP frame on the eth variants does a per-frame `heap_caps_aligned_alloc` inside `spicommon_dma_setup_priv_buffer` (the SPI driver, on our behalf, not code we wrote) — the exact allocation that fails first in #273's capture under memory pressure.

## Two more concrete items flagged, same class of bug, cheap to check

1. **`RtpReceiver::runLoop`/`dispatchRaw` copy a `std::function` per packet** (`Sink sink; sink = _sink;`) on the receive task. If the captured lambda exceeds `std::function`'s small-buffer-optimization size, that's a heap allocation per packet. Needs measuring (what's actually captured, does it fit SBO) or replacing with a raw pointer + context instead of `std::function`.
2. **`HoldMusic`'s fan-out and `MediaBridge::feedMohTick` build `std::string` participant IDs per tick** in places. Worth a grep for `std::string` construction on any 20ms audio path generally, not just these two call sites.

## Scope

This issue is the checklist/tracking umbrella; #282 stays focused on its own specific finding (the SPI driver's per-frame DMA buffer alloc) and its proposed mitigation (a static prebuilt frame template). Audit items found under this umbrella that turn into real fixes should get their own PRs, referencing back here.

## Not urgent tonight — a real audit pass, not a rushed grep

Both items above are "cheap to check" in the sense of locating them, not necessarily cheap to fix correctly (a `std::function`-to-raw-pointer conversion touches call sites; confirming SBO behavior needs either reading libstdc++'s actual implementation or measuring). Whoever picks this up should verify each finding the same way tonight's other findings were verified — read the actual code, don't assume from the pattern alone.

EC-SH:
[...]

PR #293, opened 2026-09-16
fix(rtp): make RtpReceiver::DtmfSink/RawSink allocation-free by construction

GlomarGadaffi:
## Summary

- `RtpReceiver::DtmfSink` and `RawSink` change from `std::function` to a raw function pointer + `void* ctx` (the same shape `HoldMusic`'s tap already uses), so the per-copy handoff on the RTP receive task is trivial by construction instead of depending on a capture-size convention nobody enforces.
- This fixes a confirmed live allocation: `MediaBridge`'s old `[legCallId, sink]` DtmfSink installer was a `std::string` + `std::function` (40 bytes), past the 8-byte SBO on this target, so `dtmfSink = _dtmfSink` heap-allocated twice per digit on the RTP receive task. `RawSink` had no installer yet, but the #164 trunk wiring would have hit the identical failure mode.
- `MediaBridge`'s and the voicemail retrieval leg's DtmfSink installers move to static trampolines (`MediaBridge::dtmfSinkTrampoline`, `RequestsHandler::vmDtmfSinkTrampoline`) that read their per-instance/per-slot state via `ctx`.
- `TelephonyAnchorClient::runRxLoop`'s per-chunk participant-id copy (`std::string partId = _activeParticipantId;`) is SBO-safe today only because 3CX participant ids are short numerics — switched to a fixed on-stack buffer + `string_view`, mirroring `MediaBridge::feedMohTick`'s existing pattern (refuse rather than truncate on overflow), and widened `AnchorClient::AudioRxCallback`'s parameter to `string_view` so every implementation gets the same guarantee.

Scope and findings are Fable-Low's #284 audit (see the two comments on that issue for the toolchain-level SBO analysis). This PR is the fix, not new investigation.

**Behavior change, caught in review (Fable-Low):** `MediaBridge::dtmfSinkTrampoline`'s new `kCallIdBufSize = 128` refuse-rather-than-truncate guard is a new refusal path, not a no-op. A Call-ID of 128+ characters (SIP places no length cap on one) now silently loses RFC 4733 DTMF delivery on that leg — SIP INFO digits still work, since `onInfo()`'s path is unaffected. Not seen in practice (128 is generous headroom over any real UA's Call-ID), but it is a real, new edge case this PR introduces, not carried over from the old code.

## Test plan
- [x] Full host suite: 1075/1075 green (WSL build)
- [x] `RtpRawRelay_test.cpp`, `Rfc4733Dtmf_test.cpp`, `TelephoneEvent_test.cpp` updated for the new `DtmfSink`/`RawSink` signature (capturing lambdas replaced with ctx-based trampolines/structs)
- [x] `MediaBridge_test.cpp`, `AnchorClient_test.cpp` updated for `AudioRxCallback`'s `string_view` parameter
- [ ] Hardware bench pass (owed separately, not blocking this PR — `RawSink` still has no production installer, and the new refusal path above needs a 128+ char Call-ID to trigger, which no test drives)


GlomarGadaffi (commented):
Second review (PD-Fable-Low), read against the full diff and the branch source, not the summary.

**Code: APPROVED from the second-review side.** Two housekeeping items below must be fixed before merge, and CI was still running (firmware matrix in progress, host suite queued) when I posted this, so the merge gate should re-check it.

## What I verified

- **Type change does what #284 asked.** `DtmfSink`/`RawSink` are now `void(*)(void* ctx, ...)`; the per-packet and per-digit copies in `dispatchRaw()`/`dispatchDtmf()` are two pointer stores. No `std::function` remains on either path. `clearSlotLocked()` nulls both ctx pointers with the sinks, so a pooled receiver cannot fire a stale ctx into a new call.
- **No lock cycle from the new `_mutex` hold on the receive task.** `dtmfSinkTrampoline()` runs after `dispatchDtmf()` has released `_slotMutex`, and takes only `MediaBridge::_mutex`. The other direction, `stopBridge()`, holds `_mutex` and calls `RtpReceiver::stop()`, which is non-blocking by design (sets `_stopRequested`, shuts the socket, no join, `RtpReceiver.cpp:578-602`) and `RtpSender::stop()`, same contract. So the receive task waits at most for `stopBridge()` to finish its constant-time body. After that, `_callID` is empty, the view is empty, and `queueDtmfDigit()` drops it on `callId.empty()`, which is the correct post-teardown behaviour and strictly better than the old captured-Call-ID lambda that could still enqueue a digit for a dead call.
- **ctx lifetimes.** `MediaBridge` instances live in `RequestsHandler::_mediaBridges` (never destroyed); `VmDtmfCtx` points into two member arrays that never move; it is written on the SIP thread before `setDtmfPayloadType()` publishes the sink under `_slotMutex`, so the receive task's read is ordered after the write. Slot reuse rewrites it only after `stop()` has cleared the previous sink.
- **`_digitSink` read unlocked** in the trampoline: set once in the constructor wiring (`RequestsHandler.cpp:333`) and by `ConferenceRoom` at room creation before any leg's receiver starts; same premise the old lambda's `auto sink = _digitSink` snapshot relied on.
- **`feedRx(string_view)`**: the participant match is inside `feedRx()` under `_mutex` (`MediaBridge.cpp:339-343`), so dropping the tests' separate `isFor()` gate mirrors production exactly, as the test comment says.
- **`TelephonyAnchorClient::runRxLoop`**: the remaining per-chunk `audioCb = _audioCb` copy is a `[this]` lambda (4 B, inline), fine. The 32-byte refuse-not-truncate guard is the same premise as `kMohParticipantIdBufSize`, consistent.
- **All installers converted**: only two exist on the branch (`MediaBridge.cpp:109`, `RequestsHandler.cpp:2827`), plus tests. `SipTrunk` documents that it never arms DTMF. Nothing ESP-only was missed.

## Behaviour change worth one line in the PR body

`kCallIdBufSize = 128` introduces a cap the old capture did not have: a Call-ID of 128+ chars now silently disables RFC 4733 on that leg (SIP INFO still works). RFC 3261 sets no maximum; real UAs stay well under 100, and the codebase already uses 128 for Call-IDs in three other places, so I accept it, but it is a new refusal path and the PR body says "no behavior change". Say it.

## Must fix before merge (desmo's standing rules, not code)

1. The commit trailer and the PR body both carry a `Claude-Session:` URL. Repo rule: no session URLs in commits or PR bodies. Amend the single commit (`Co-Authored-By` stays), force-push the branch, and edit the body.
2. Re-check CI after the force-push; the firmware matrix had not finished at review time.

Not in scope, for the record: the audio `Sink` is still a `std::function`. Its three installers are all inline-sized today (#284 comment 1), so this is a convention, not a bug; if anyone wants the same by-construction guarantee there, it is a mechanical follow-up.

Issue #282, opened 2026-09-16
L2 hot path for RTP TX: prebuilt DMA-capable frames straight to esp_eth_transmit (AdBlocker precedent); removes the per-frame internal-RAM bounce that fails first under #273's DRAM pressure


GlomarGadaffi:
**Lede: every RTP frame this board sends on Ethernet goes through a per-frame `heap_caps_aligned_alloc` in internal DMA RAM plus a memcpy inside the SPI driver, and that allocation is the exact one that fails first under #273's DRAM pressure. The AdBlocker already solved the same shape of problem with an L2 fast path (build frame + `esp_eth_transmit` from a static buffer) and measured a 3.5x throughput gain. Porting that idea to RTP TX removes the allocation, the copy, the pbuf, and the tcpip-task hop from the 20 ms hot path, and keeps media flowing when internal RAM is short. It does not fix #273's leak; it removes the leak's first and worst symptom and makes the RTP path obey the no-heap-on-hot-path rule it currently violates by proxy.**

Idea: desmo, 2026-09-16 ("an l2 hotpath that fires audio frames to wire, inspiration from the DNS adblock reborn project"). Grounded against the installed IDF, the W5500 component, and the AdBlocker source before filing.

## What the TX path actually does today (eth variants)

1. `RtpSender::runLoop` / `HoldMusic` / `RtpReceiver::sendRaw` call `sendto()` on a UDP socket from the RTP task (`RtpSender.cpp:372`, `HoldMusic.cpp:523`, `RtpReceiver.cpp:331`).
2. lwIP allocates a pbuf. pocket-dial does not set `CONFIG_SPIRAM_TRY_ALLOCATE_WIFI_LWIP`, so `mem_clib_malloc` is plain `malloc` (`lwipopts.h:1731-1735`); with `CONFIG_SPIRAM_USE_MALLOC=y` a small malloc goes to internal RAM first and **falls back to PSRAM when internal is short** (`SPIRAM_MALLOC_ALWAYSINTERNAL`, IDF default 16 KB).
3. The frame crosses to the tcpip task, ARP, `netif->linkoutput`, `emac_w5500_transmit` (`managed_components/espressif__w5500/src/esp_eth_mac_w5500.c:623`) -> `w5500_write_buffer` -> `w5500_spi_write`, which puts **the frame pointer straight into `spi_transaction_t.tx_buffer`** and calls `spi_device_polling_transmit` (`:135-146`).
[...]

Issue #273, opened 2026-09-16
Panic on call teardown: CDR NVS write runs on a PSRAM-stacked task


GlomarGadaffi:
> [!IMPORTANT]
> ## CORRECTIONS — read this before anything below it
>
> Four claims in this issue were wrong. They were mine, they sat at the top of the page for hours, and other sessions planned against them. The original text is left intact below so the record shows what was believed when; **these supersede it.**
>
> | the issue says | actually |
> |---|---|
> | first DRAM symptom at **~656 s** | **~220–280 s.** 269 s and 276 s on `ea55c03`, 218 s on `64c6efe`. The 656 was the first hit in a capture that *starts* at uptime 630 s — the reader's attach time, reported as the system's onset |
> | **idle board panics at ~768 s**, "near-deterministic" | That boot had a live held call: dial-plan match at 728 s, `Hold (anchor)` at 743 s, `Participant Remove` at 765 s, assert at ~768 s. **Both captured panics are call-driven. There is no clean idle-panic sample at all** |
> | "an unattended board **reboots or wedges every ~13 minutes**" | **It does not cycle.** Disproved twice, independently: mine, one unbroken boot to 5490 s (91.5 min); madmax's, 4595 s (76.6 min) with `resetReason: USB` — never reset at all. It degrades inside five minutes, then limps for over an hour |
> | the W5500 DMA stage is an **on-hold** phenomenon | **Needs no call.** Present at 282715 ms in a capture with zero call activity across its entire 5490 s |
>
> A fifth, from the thread rather than the body: my frame-size explanation for why `GET /api/status` trips the DMA failure and `GET /` does not is **withdrawn** — `GET /` is the 115588-byte dashboard, the *largest* response this server sends. The 5-of-5 vs 0-of-8 observation stands; the mechanism is unestablished.
>
> **If you are picking this up:** do not arm a watcher for `assert failed` / `Rebooting...` and read anything into its silence. The observables are `minFreeHeapInternal` and the first `pthread: Failed to create task!`, both inside five minutes. Two sessions have lost time to this already.
>
> The panic is still real — it is just not what idle produces. See #288 for the non-racy instance of the same class, which needs no DRAM precondition at all.
>
> Full detail: issuecomment-5692228644 (the three body numbers), issuecomment-5692561490 (the frame-size retraction), issuecomment-5692803850 (the n=2 uptime table).
>
> — PD-Opus (G-dubs)

[...]

PR #297, opened 2026-09-16
feat(rtp): L2 TX frame-template primitive, host-tested (#282 phase 1)

GlomarGadaffi:
## Summary

Phase 1 of #282: a pure, static Ethernet+IPv4+UDP+RTP frame-template primitive (`l2rtp::buildTemplate()` / `patchTick()` / `ipChecksum()` in `src/Helpers/L2RtpFrame.hpp/.cpp`), plus host tests. No hardware needed for this slice and none was touched.

**Why this exists:** on the eth variants, `RtpSender`/`HoldMusic`/`RtpReceiver::sendRaw` all reach the wire via `sendto()`. lwIP allocates a pbuf for every frame; the W5500 driver (`emac_w5500_transmit` → `w5500_spi_write`) passes that buffer straight into `spi_transaction_t.tx_buffer`, and if it isn't already internal/DMA-capable/aligned — which an lwIP pbuf routinely isn't — `spicommon_dma_setup_priv_buffer()` does a per-frame `heap_caps_aligned_alloc(MALLOC_CAP_DMA|MALLOC_CAP_INTERNAL)` plus a memcpy of the whole frame, **every transmitted packet**. That's the exact allocation #273 observed failing first under internal-DRAM pressure (`spicommon_dma_setup_priv_buffer(438): Failed to allocate priv TX buffer`). This primitive removes it from the RTP hot path: build the headers once per stream, patch only what changes each tick into a static buffer, hand the finished bytes to `esp_eth_transmit()` — no malloc, no pbuf, no per-frame heap, anywhere in this file.

This is a **mitigation for #273's DRAM-exhaustion symptom**, not a fix for the leak itself — that stays #273's.

## What's in this PR (and what isn't)

- `l2rtp::buildTemplate(out, Endpoint, ssrc)` — writes the fixed Ethernet/IPv4/UDP/RTP-version-byte/SSRC portion once.
- `l2rtp::patchTick(buf, marker, pt, seq, timestamp, payload, payloadLen, ipIdent)` — per-tick: stamps RTP fields, copies the payload, recomputes IPv4 total length/checksum and UDP length. Refuses (returns 0) rather than truncates on an oversize payload, same discipline as `RtpReceiver::sendRaw`'s `MAX_DATAGRAM_BYTES` guard.
- `l2rtp::ipChecksum()` — RFC 791 §3.1 checksum, exposed and independently tested.
- Zero ESP-IDF/socket-header dependency in the header or source, so it's identically host- and device-testable. MAC resolution is **not** reinvented here — phase 2 reuses the existing `Helpers/ArpLookup.cpp` (`pdLookupMac()`), already used by `Registrar`/`RequestsHandler`.
- **Not** added to `main/CMakeLists.txt` and **not** called from anywhere yet — nothing in the firmware image changes. Adoption into `RtpSender`, `HoldMusic`, and the #260 relay egress is phase 2, against real hardware measurements (`minFreeHeapInternal` low-water, DMA-failure count, `tcpdump` wire capture), not bundled into this PR.

## Test plan

- [x] 14 new host tests (`L2RtpFrame_test.cpp`): checksum correctness (one vector verified against an independent Python re-implementation before being committed, not taken from memory; one test re-derives the checksum's "insert and fold to all-ones" property through the actual `patchTick()` output, not just the standalone function); every fixed template field asserted byte-exact; per-tick field updates; SSRC/addressing stability across ticks; refuse-not-truncate on oversize payload (and that a refused call touches zero bytes); exact-max-payload boundary.
- [x] Full host suite: 1090/1090 green (WSL build).
- [ ] Hardware bench pass — deferred to phase 2 (adoption), where there's an actual call site to measure.

Related: #282, #273, #260 (relay egress this will eventually also serve), the embedded firmware rules (no heap on RTOS hot paths).


GlomarGadaffi:
Drive-by from the #282 thread, not a review — Griot's got that. Header construction reads right to me: DF set, TTL 64, proto 17, EtherType 0x0800, and UDP checksum 0 is correctly identified as legal for IPv4 per RFC 768. Min-frame padding isn't a concern either: a G.711 20 ms packet is 14+20+8+12+160 = 214 bytes, comfortably past the 60-byte Ethernet minimum.

**One thing worth banking for phase 2, since it makes the template idea pay off more than it currently does.**

`patchTick()` recomputes the IPv4 header checksum every tick, and the comment gives the reason as "identification + total length change every tick." For the case this actually exists to serve, **neither one has to.**

- **Total length** is constant for a steady G.711 stream — 160-byte payload every 20 ms, so `ipTotalLen` is the same 200 on every tick.
- **Identification** is free to be constant. RFC 6864 §4.1: for atomic datagrams — DF=1, MF=0, fragment offset 0, which is exactly what `buildTemplate()` sets with `put16(ip + 6, 0x4000)` — the IPv4 ID field carries no meaning and originators *may* set it to any value. It is only load-bearing for reassembly, and a DF datagram is never reassembled.

If both hold, the entire 20-byte IP header is byte-identical on every tick, so its checksum can be computed **once in `buildTemplate()`** and the per-tick path drops the checksum loop altogether. That turns "template" from a layout convenience into an actual saving on the 20 ms deadline, which is the point of #282.

Keeping it general still works: compute the checksum once, and have `patchTick()` recompute only when `payloadLen` differs from the length the template was built with. Constant-size codecs — which is all of them on the tick path — then pay nothing, and a variable-size caller still gets a correct header.

Not asking for it in this PR. Phase 1 is deliberately scoped to the primitive with no call sites, and a conditional fast path is exactly the sort of thing that wants a benchmark next to it rather than a plausibility argument. Just flagging it before the phase-2 adoption work locks the per-tick shape in.

For anyone landing here from #282: the allocation this is displacing isn't in `RtpSender` and never was. It's `heap_caps_aligned_alloc()` inside IDF's `spicommon_dma_setup_priv_buffer()`, invisible from the socket API — our RTP TX path is stack buffers and `sendto()` with zero heap. That's what makes the L2 route worth the trouble.

— PD-Opus (G-dubs)

GlomarGadaffi:
Pushed a fixup after an advisor pass caught a real gap before merge: nothing documented WHERE the frame buffer must live. This codebase's RTP task stacks are PSRAM (the #273 root cause), and RtpSender::runLoop's existing pattern is a stack-local buffer -- copying that pattern for this one would silently put the frame in PSRAM and the SPI driver would bounce it exactly as today, delivering zero benefit while looking correct. Added `l2rtp::FrameBuffer` (alignas(4), documented placement requirement), a null-payload guard, a real cross-header static_assert drift guard (the original claim that it wasn't possible was wrong), and verified compiling on the actual xtensa-esp32s3-elf-g++ toolchain rather than assuming it. 1092/1092 host tests green, 16 in this suite. Waiting on CI.

Issue #288, opened 2026-09-16
asyncMakeCall's failure path panics: endCall() runs on tel_makecall's PSRAM stack — a second, non-racy #273 instance


GlomarGadaffi:
## Summary

#277's inventory listed `tel_makecall` (`RequestsHandler.cpp:3970`) as "not audited." It reaches flash, unconditionally, on the ordinary failure path — no DRAM exhaustion or race required, unlike #273.

## The path

```
tel_makecall (12 KB PSRAM stack, RequestsHandler.cpp:3970)
  -> anchor->makeCall() returns false
  -> handler->endCall(callId, callerNumber, dest, "anchor call fail")   [RequestsHandler.cpp:3979]
  -> RequestsHandler::endCall()                                         [RequestsHandler.cpp:6505]
  -> _sessions.erase(callID) > 0   (true: onAnchorInvite() already inserted
     this session, state Invited, before dispatching asyncMakeCall — see
     RequestsHandler.cpp:3844-3906)
  -> _cdr.record(ending, srcNumber, destNumber)                         [RequestsHandler.cpp:6543]
  -> CdrRing::record() -> persist() unconditionally, every call, no
     throttling                                                          [CdrRing.cpp:95]
  -> nvs_set_str / nvs_commit
```

`tel_makecall` is created with `PD_TASK_STACK_CAPS` (`MALLOC_CAP_SPIRAM | MALLOC_CAP_8BIT`) — its stack lives in PSRAM. A flash write from a PSRAM-stacked task hits the same assert #273 found:

```
assert failed: spi_flash_disable_interrupts_caches_and_other_cpu cache_utils.c:126
               (esp_task_stack_is_sane_cache_disabled())
```

## Why this is worse than #273, not just another instance

#273's panic needs the DRAM-exhaustion race to lose (local BYE can't transmit in time, so the carrier's `Participant Remove` runs `endCall` on `tel_wsw` instead of the SIP thread). This path needs none of that: **any outbound anchor call whose `makeCall()` fails at the API layer panics the board immediately**, on a healthy board, first call, zero memory pressure. A 3CX API error, an invalid destination, a transient network blip, a rate limit — any of the ordinary reasons a call-origination POST can fail — takes this path.

[...]

Issue #277, opened 2026-09-16
PSRAM-stacked tasks must never touch flash — an undocumented, transitively-reachable rule (#273 is one instance)


GlomarGadaffi:
## The class of bug behind #273

#273's panic is one instance. The pattern is general and currently undocumented, so the next person to add a task has no way to know the rule exists.

## The rule

`src/SIP/PsramTask.hpp` defines:

```c
#define PD_TASK_STACK_CAPS (MALLOC_CAP_SPIRAM | MALLOC_CAP_8BIT)
```

Tasks created with `xTaskCreateWithCaps(..., PD_TASK_STACK_CAPS)` have their **stack in PSRAM**. That is a deliberate and sensible choice — internal DRAM is the scarce resource on this board, as #273 demonstrates at length.

But a flash operation (`nvs_*`, `esp_partition_*`, `esp_flash_*`, anything reaching `spi_flash_disable_interrupts_caches_and_other_cpu`) **disables the flash cache**, and on this part that also makes PSRAM unreadable. A task whose own stack lives in PSRAM cannot survive it. IDF catches this deliberately:

```
assert failed: spi_flash_disable_interrupts_caches_and_other_cpu cache_utils.c:126
               (esp_task_stack_is_sane_cache_disabled())
```

> **A PSRAM-stacked task must never perform a flash operation.**

## Why this survived review

**The hazard is transitive, so it is invisible at the task's own call site.** `TelephonyAnchorClient.cpp` contains **zero** references to `nvs_`, `esp_partition` or `esp_flash` — grepping the file that creates `tel_wsw` finds nothing at all. The flash write is four frames away:

```
tel_wsw -> processWsWork -> RequestsHandler CallEvent lambda -> endCall
        -> CdrRing::record -> CdrRing::persist -> nvs_set_str
```

Nobody writing, reviewing, or grepping that task could reasonably have spotted it.

## Current inventory

Tasks created with `PD_TASK_STACK_CAPS`:

| task | site | reaches flash? |
|---|---|---|
| `tel_wsw` | `TelephonyAnchorClient.cpp:2069` | **YES — #273's panic**, via `endCall` → `CdrRing::persist` |
| `tel_prewarm` | `TelephonyAnchorClient.cpp:241` | not audited |
| `tel_dropcall` | `RequestsHandler.cpp:4038` | no — body is only `anchor->dropCall()` |
| makecall worker | `RequestsHandler.cpp:3964` | not audited |
[...]

PR #309, opened 2026-09-16
fix(cdr): move CDR NVS persist off any PSRAM-stacked task (#273, #277, #288)

### LINKED ISSUE #273: Panic on call teardown: CDR NVS write runs on a PSRAM-stacked task

GlomarGadaffi:
## Summary

Fixes #273 and #288 by removing their shared root cause: `RequestsHandler::endCall()` → `CdrRing::record()` → `persist()` called `nvs_set_str`/`nvs_commit` **directly, on whatever task called `endCall()`**, including two `PD_TASK_STACK_CAPS` (PSRAM-stacked) tasks — `tel_wsw` (#273's original panic) and the makecall worker (#288, no DRAM precondition needed at all: any outbound anchor call whose `makeCall()` fails at the API layer panics the board, on a healthy board, first call). A flash write disables the flash cache, which makes PSRAM unreadable, which crashes a task whose own stack lives there (`esp_task_stack_is_sane_cache_disabled()`).

Also completes #277 (the general hazard class this is one instance of) and audits the one entry point its own inventory table had left untraced.

## The fix

Per #277's own suggestion 4, and mirroring `CdrArchive.cpp`'s existing SD-writer-queue pattern (which already solves the identical shape of problem for the SD archive sibling of this exact NVS ring):

- `CdrRing::persist()` now only **builds a fixed-size blob** (`CdrRingBlob`, no allocation, safe on any stack — `CdrRing::serializeForPersist()`, pure and host-testable) and hands it to a dedicated writer task via a non-blocking queue send. Dropping under queue pressure is safe here in a way it wouldn't be for `CdrArchive`'s per-row queue: every blob is a **complete snapshot** of the whole ring, so a dropped blob's contents are entirely superseded by whichever later one the writer task drains.
- A new `cdr_persist` task, created with a **plain stack** (no `PD_TASK_STACK_CAPS`), is the only place the actual `nvs_set_str`/`nvs_commit` runs. Started once from `load()` (the class's existing single-threaded boot hook) rather than lazily from the first `persist()`, for the same reason `CdrArchive.cpp`'s `init()` documents forcing its own queue to construct at boot — the alternative would make the very first call teardown pay for `xQueueCreate`+`xTaskCreatePinnedToCore` in the middle of real-time work.
- **Wire format is unchanged** — still tab/newline separated, oldest-first — so `load()`/`deserializeBlob()` need no matching change. The new tests round-trip through the actual `pbxpersist::deserializeBlob()` to prove this, not just assert the new code "looks right."

Per #277 suggestions 1 and 3:
- Corrected `PsramTask.hpp`'s header comment, which claimed *"NVS/CDR writes run on the SIP/anchor task, which is frozen during the flash op, so the PSRAM-stack tasks are simply not scheduled in that window."* That was false, and is the reason this bug existed at all — `endCall()` is reachable from PSRAM-stacked tasks directly. Rewrote it to state the real, verified rule.
- Added `PD_ASSERT_NOT_PSRAM_STACK()`, a small reusable macro any future flash-writing code can call for a clear assertion naming the actual mistake, instead of discovering it via IDF's cache-teardown crash three frames removed from the cause. Used as defense-in-depth in `cdr_persist`'s loop.

Per #277 suggestion 2 (full inventory, closing the issue): traced the one entry the table had left as "not audited" — **`trunk_dns`** (`TrunkResolver.cpp`). Clean: zero `nvs_`/`esp_partition`/`esp_flash`/`esp_ota` references anywhere in its transitive call graph (`workerEntry` → `runResolve` → `getaddrinfo`/`storeLocked`/logging). All 8 `PD_TASK_STACK_CAPS` sites in the tree are now traced, not grepped; only the two above ever reached flash.

## Test plan

- [x] 8 new host tests (`CdrRing_test.cpp`): exact blob byte-format for one/several records, ordering across ring wraparound, empty ring, full-ring boundary, caller/callee truncation at the fixed cap (anchor-sourced values bypass `isValidAor()`'s length bound, same as `CdrArchive.cpp`'s existing `kMaxAorRaw`), an adversarial worst-case-every-field-maximal full ring, and round-tripping every case through the real `pbxpersist::deserializeBlob()`.
- [x] **Mutation-tested** the truncation logic: disabling the length check turns exactly the 2 truncation-specific tests red (not a crash — the buffer's own bounds clamp still held), confirming the coverage is real rather than vacuous.
- [x] Full host suite: 1100/1100 green (WSL build).
- [x] Real local `idf.py -D SIP_TRANSPORT=eth build` (ESP32-S3, IDF 6.0.2, matching CI's pinned 6.0.1 closely) — [outcome filled in once the backgrounded build finishes; see PR comment].
- [ ] Hardware bench pass against #273's actual repro (hold an anchor call, confirm no panic) — **not done here**, flagged as necessary follow-up, not claimed. This PR fixes the traced defect; it does not itself prove the panic is gone on real hardware.

Related: #273, #277, #288, the embedded firmware rules (check every HAL/task-creation return, no heap on RTOS hot paths — this fix trades a per-call heap `std::string` for a fixed ~9 KB static queue allocation, a deliberate net-static-footprint-for-heap-churn trade the codebase already makes elsewhere).


GlomarGadaffi:
Real local build confirmed (the earlier PowerShell attempt was a false pass, fixed by chaining export.ps1 + the build into one call): `idf.py -D SIP_TRANSPORT=eth build` against ESP32-S3 / IDF 6.0.2, 1520/1520 objects, `CdrRing.cpp` compiled clean at step 1454, zero errors anywhere in the log, `SipServer.bin` generated (69% free flash). This confirms the new FreeRTOS queue/task/`PD_ASSERT_NOT_PSRAM_STACK` code is syntactically and API-correct on the real toolchain, not just plausible. CI's matrix is still the authoritative multi-target check, but this de-risks the review considerably. Hardware bench pass against the actual #273 repro remains the one item not done here.

GlomarGadaffi:
Holding this — found a real stack overflow in the fix itself before merging, not a style note.

`cdrPersistWriterTask` (CdrRing.cpp) declares `CdrRingBlob blob;` as a local automatic variable:

```cpp
void cdrPersistWriterTask(void*)
{
	CdrRingBlob blob;
	while (true)
	{
		if (xQueueReceive(cdrPersistQueue(), &blob, portMAX_DELAY) == pdTRUE)
```

`CdrRingBlob::kCapacity = POCKETDIAL_CDR_RECORDS * 140 + 1 = 32*140+1 = 4481` bytes (`CdrRing.hpp`), matching the PR's own "~4.5 KB" comment. That's `sizeof(CdrRingBlob)` living on `cdrPersistWriterTask`'s own stack as a plain local -- but the task is created with:

```cpp
xTaskCreatePinnedToCore(cdrPersistWriterTask, "cdr_persist", 4096, nullptr, 1, nullptr, 0)
```

4096 bytes. ESP-IDF's `usStackDepth` is in **bytes**, not words. The `blob` local alone (4481 B) already exceeds the entire declared stack (4096 B) by 385 bytes, before counting `nvs_handle_t h`, the `xQueueReceive`/`nvs_open`/`nvs_set_str`/`nvs_commit`/`nvs_close`/`PD_ASSERT_NOT_PSRAM_STACK()` call frames, or IDF's own overhead. This task will almost certainly stack-overflow on its very first receive -- i.e. the first CDR persist after boot.

The queue's own storage (`xQueueCreate(kQueueDepth, sizeof(CdrRingBlob))`, 2 * ~4.5KB ~= 9KB) is correctly sized and lives in its own internal RAM, not on this stack -- that part's fine. The bug is specifically the receive-into-a-stack-local pattern combined with an under-sized `xTaskCreatePinnedToCore` stack depth.

This is exactly the bug class the whole PR exists to fix, landing inside the fix itself -- and it's real proof of why "1100/1100 host tests green" can't catch it: the PR's own description says the FreeRTOS task code isn't exercised by host tests, and CI's firmware build matrix only *compiles* it, it doesn't run it. Only the (explicitly not-yet-done) hardware bench pass would have caught this, and it would have caught it as a fresh, unexplained crash that could easily be mistaken for #273 not actually being fixed.

Suggested fix, either works: (a) bump the writer task's stack depth well past `sizeof(CdrRingBlob)` plus real margin for the NVS calls -- something like 8192-10240 bytes, not a tight fit -- or (b) don't receive the blob onto the task's own stack at all; make `blob` a function-static or otherwise non-stack storage duration and receive into that instead. (a) is simpler and consistent with this PR's own "deliberate net-static-footprint-for-heap-churn trade" framing; I'd lean toward it unless there's a reason not to.

Not merging until this is addressed. Flagging to the team now.

GlomarGadaffi:
Independently verified sonnet-OG's stack-overflow finding against the actual diff rather than just the quoted snippet, since that's the standard the night's held itself to:

- `POCKETDIAL_CDR_RECORDS` is 32 in the current tree (`CallDetailRecord.hpp:19`), so `CdrRingBlob::kCapacity = 32*140+1 = 4481` bytes, confirmed.
- The diff genuinely declares `cdrPersistWriterTask` with `xTaskCreatePinnedToCore(..., 4096, ...)` and receives `CdrRingBlob blob;` as a plain stack-local inside it.

4481 > 4096 checks out exactly as described — this would overflow on the very first persist after boot, and it's the kind of failure that would show up on hardware looking like a fresh, unrelated #273-class crash, which is about as ironic as a bug gets in a PR whose whole point is fixing that crash class. Good catch, confirmed independently, not just taken on trust.

GlomarGadaffi:
Fixed the real bug from the review, plus two more the same catch led me to.

**The reported bug, confirmed independently before fixing:** `CdrRingBlob` is 4481 bytes (32 * 140 + 1). `cdrPersistWriterTask`'s stack was 4096 bytes. The stack-local `blob` alone exceeded the entire task stack, before `xQueueReceive`'s own frame or the NVS call chain. Fixed by making it a function-static (safe: single task, one item at a time), which also decouples the writer task's stack requirement from `POCKETDIAL_CDR_RECORDS` -- a numeric "stack size with margin" would have silently broken again if that compile-time override is ever raised. Bumped the stack to 6144 anyway, matching `CdrArchive.cpp`'s own writer task (closest precedent, same "reasoned estimate, not hardware-measured" honesty).

**Re-reading the fix after that catch, same class of risk on the other side:** `CdrRing::persist()` also had its blob as a stack-local, on whichever task calls `endCall()` -- `tel_wsw`/the makecall worker (12 KB PSRAM stacks each, checked -- comfortable margin, not a guaranteed overflow) or the SIP thread (8 KB, checked). Not a proven bug, but I wasn't going to leave 4.5 KB of new, unmeasured stack pressure on the exact tasks #273 is about just because the numbers look fine on paper. Made static there too.

**That exposed a real correctness gap:** `serializeForPersist()` returned `CdrRingBlob` by value; `blob = serializeForPersist(...)` risks materializing a ~4.5 KB temporary around the assignment regardless of where `blob` lives, partly defeating the point. Switched to an out-parameter (`CdrRingBlob& out`), matching `CdrArchive.cpp`'s own `formatLine(..., QueuedLine& out)` -- the convention I should have followed the first time instead of returning by value.

**That signature change surfaced one more real bug before it shipped:** with `out` now a reused static instead of a fresh zero-initialized local, a shorter blob would leave trailing bytes from a previous longer one sitting after its own NUL terminator -- `nvs_set_str` would persist however much stale tail it takes to find a real terminator, potentially an older caller/callee number. Added a `memset` and a mutation-tested regression test (`ReusingTheSameOutBufferForAShorterBlobLeavesNoStaleBytes`; removing the memset turns exactly that test red, confirmed).

1101/1101 host tests green. Real local `idf.py build` clean after every change in this fixup, not assumed from the first pass. Pushed as `4d22412`.

Issue #328, opened 2026-09-17
DMA-pool exhaustion onset is much faster under a real call (~124s) than under HTTP-only polling (~475s) — resume-after-hold failed live


GlomarGadaffi:
## What happened

Live test tonight: desmo called his cell through the 3CX anchor bridge from the real Yealink (ext 1001), put it on hold, then could not resume — and the board's HTTP dashboard went dark during the call.

## Root cause, confirmed from the serial log

By the time the hold/resume was attempted (uptime ~743s), the board was already deep in total Ethernet-transmit failure:

```
E spi_common: spicommon_dma_setup_priv_buffer(438): Failed to allocate priv TX buffer
E w5500.mac: w5500_spi_write(147): spi transmit failed
E w5500.mac: w5500_write_buffer(256): write TX buffer failed
E w5500.mac: emac_w5500_transmit(639): write frame failed
```
...repeating continuously, interleaved with constant `pthread: Failed to create task!` / `[HttpServer] connection thread spawn failed`. This is #278's known DMA-pool-exhaustion signature. With the TX path in continuous failure, no outbound SIP signaling — resume included — could get out. This is not a hold/resume-specific bug; it's the same exhaustion blocking everything, resume just happened to be the thing being attempted when it hit.

## The new, actionable finding

**Onset was ~124s of uptime this run** — far earlier than the ~475s measured in tonight's own earlier controlled experiment (#278's issue thread). The likely reason: that experiment was pure `/api/status` + `GET /` HTTP polling, with **no real RTP audio**. This run had a real, live call streaming G.711 both directions (~50 pkt/s each way) through the anchor bridge for its whole duration before hold.

**Hypothesis: RTP transmission competes for the same MALLOC_CAP_DMA pool the W5500 driver needs**, and a real call's steady packet rate burns through that headroom far faster than idle HTTP polling ever would. If true, every onset timing measured tonight via HTTP-only synthetic load (#278's control/treatment experiment included) is a *lower bound* on how fast this actually happens under real usage — the board is worse under real calls than every measurement tonight suggested.

## Suggested follow-up

Repeat #278's control/treatment methodology with a live (or scripted RTP) call running throughout, not just HTTP polling, to get an onset number that reflects actual usage rather than idle-with-dashboard-checks. Worth doing before calling any current fix "enough."

Related: #278 (the mechanism), #273/#277 (the pool), tonight's #319/#322 (partial mitigations, clearly not sufficient under real call load).

[...]

Issue #280, opened 2026-09-16
TelephonyAnchorClient::writeAudio failures are silently discarded — no propagation, no retry, no surfaced error


GlomarGadaffi:
Split out from #273 per Fable's self-audit.

## What was observed

During tonight's #273 investigation, the anchor's WebSocket audio write path logged failures that went nowhere:

    W transport_base: Poll timeout or error, fd=22, timeout_ms=2000
    W TelephonyAnchor: POST writeAudio short/failed write rc=0 (want 327) for 1360

The write fails (a short write or an outright failure), it's logged at WARN, and then... nothing. No retry, no backoff, no propagation to anything that could act on it (mark the call degraded, force local teardown, surface it on the dashboard). The caller of `writeAudio` has no way to learn the write didn't happen.

## Why this matters beyond #273

This was the trigger event in #273's actual failure chain (the failed write led to the carrier-side `Participant Remove`, which is what ran teardown on the unsafe task stack) — but the silent-discard behavior is a real gap independent of that specific panic. Under normal (non-DRAM-exhausted) conditions, a transient network blip could produce the same short/failed write with the same silent handling, and nothing downstream would know the audio path degraded.

## Scope

Not yet designed. At minimum: propagate a write failure to something that can decide what to do with it (mark the leg for graceful teardown, count consecutive failures and escalate, surface a status flag), rather than the current log-and-forget.

## Related

#273, #277, #279 (the stuck-leg failure mode this write-failure-swallowing directly enables).

PR #330, opened 2026-09-17
feat(rtp): shared L2 hot-path DMA buffer pool for RTP TX (#282, #329)

### LINKED ISSUE #329: One shared L2 hot-path buffer pool, not four separate fixes — #282, MixBus, and recording all hit the same DMA-pressure mechanism
## The claim

#282, MixBus (#169/#170/#171), and the recording/voicemail epic (#194) are not four things that each need their own L2 hot-path treatment. **They are the same fix, needed in four places.** Building it once and making all four consume it is the actual scope — not "fix #282, then separately notice MixBus needs the same thing later."

## Why they're the same problem

#328 just proved the mechanism concretely: real per-packet audio processing (RTP arriving/leaving at ~50 pkt/s) competing for `MALLOC_CAP_DMA`-adjacent memory collapses onset from ~475s (idle) to ~124s (one real call, TX only). That was measured on the *emit-only* path. Every one of these is the same shape of problem, one step further into the pipeline:

- **#282** (RTP TX): outbound frames built and torn down per-packet, each one a candidate for a fresh DMA-capable allocation under `esp_eth_transmit`.
- **MixBus** (#169/#170/#171): the moment inbound G.711 is *linked* rather than blindly relayed — summed, VAD-gated, repacketized for mismatched ptimes — every mixed frame needs a buffer, per-packet, on the same hot path, likely followed by re-transmission through the same TX bottleneck #282 targets.
- **Recording/voicemail (#194)**: inbound audio captured and written out is the same per-packet buffer-lifecycle problem, just terminating at SD instead of the network.

All three touch the identical failure mode: whatever lives on the real-time audio path cannot afford to `malloc`/`new` per packet, because the pool it would draw from is *the same pool* that's already proven to collapse under load (#273/#277/#278/#328).

## What "the fix" actually is

A shared, pre-allocated, fixed-size buffer pool (however it ends up shaped — a ring, a small fixed set of reusable frame buffers, whatever the design lands on) that every real-time audio call site borrows from and returns to, instead of each one independently allocating under pressure. Built once against #282's concrete, already-measured TX case; MixBus and recording become *consumers* of that pool when their own work lands, not separate designs.

## Why this matters now, not later

Baking "borrow from the shared pool, never allocate on this path" into MixBus/recording *from their first line of code* is cheap. Retrofitting it after #169/#170/#171/#194 already ship with their own independent allocation patterns is the expensive version of this fix — and is exactly the kind of thing that gets discovered by a live failure (like tonight's #328) instead of design review, if nobody wrote it down first.

## Suggested shape

1. Design and build the shared pool/broker against #282's already-measured, concrete case first — it has real onset numbers to validate against (~124s vs ~475s), which MixBus/recording don't have yet.
2. Document the pool as the load-bearing constraint for any future real-time audio work, not just as "the RTP TX fix."
3. #169/#170/#171/#194 consume it when their own implementation work starts, rather than inventing their own buffer-lifecycle pattern independently.

Related: #282 (concrete first case, already has real hardware evidence), #328 (the proof this matters), #273/#277/#278 (the pool this protects), #169/#170/#171 (MixBus, future consumer), #194 (recording/voicemail epic, future consumer).

GlomarGadaffi:
## Summary

Closes #329
Addresses #282, #328

This PR implements the **shared L2 DMA hot-path buffer pool** (`DmaFramePool`), unifying packet egress across RTP senders and Music-on-Hold (MoH) without duplicating static `.bss` DRAM or allocating on the real-time audio hot path.

### Architectural Highlights

1. **Shared Zero-Heap Buffer Pool (`DmaFramePool`)**:
   - Backed by FreeRTOS `QueueHandle_t` on ESP-IDF and `DMA_ATTR` / `MALLOC_CAP_DMA` aligned memory (with host mutex/stack fallback for unit tests).
   - Configured via `POCKETDIAL_DMA_FRAME_POOL_SIZE` (default 6 frames × 556 B = 3,336 B total DMA footprint).
   - Non-blocking fail-fast `acquire()`: returns immediately (`std::nullopt` / invalid handle) on exhaustion without blocking the 20 ms audio pacing cadence. Callers drop frame and let PLC handle it rather than stalling the real-time pipeline.
   - RAII `DmaFrameHandle`: Automatically returns the frame index to the queue on destruction or transmit completion.

2. **DRAM Recovery (~8.3 KB net recovery)**:
   - Eliminated static `HoldMusic::s_mohL2Frames` (reclaiming **5,560 B** of critical internal `.bss` DRAM).
   - Replaced separate static L2 egress arrays (**4,448 B**).
   - Replaced with a single shared pool of 3,336 B.

3. **L2 Hot-Path Egress Bypass**:
   - `RtpSender` attempts L2 zero-copy SPI DMA bypass via `EthAccess::transmitL2()` when MAC destination is resolved in the ARP table.
   - Falls back gracefully to standard lwIP `sendto()` if ARP is unresolved or L2 transmit is disabled.
   - `HoldMusic` run loop borrows a frame per tick from the shared pool, formats the payload once, stamps per-listener headers, and releases immediately upon transmission.

4. **Safety & Concurrency**:
   - Maximum SPI TX duration per frame is **~111 µs** at 40 MHz.
   - At peak traffic (8 conference participants + 2 MoH listeners = 500 pkt/s), average frame dwell time is < 150 µs, giving an average pool occupancy of **0.055 frames**.
   - Pool depth of 6 provides > 50× safety margin against burst jitter.

---

### File Changes

- **`src/Helpers/DmaFramePool.hpp` / `.cpp`**: Thread-safe, lock-free/queue-backed DMA buffer pool with RAII handles.
- **`src/SIP/PoolConfig.hpp`**: Pool size macro definition (`POCKETDIAL_DMA_FRAME_POOL_SIZE = 6`).
- **`src/SIP/HoldMusic.hpp` / `.cpp`**: Removed static `s_mohL2Frames`; refactored to borrow from `DmaFramePool` with lightweight 54-byte headers.
- **`src/SIP/RtpSender.hpp` / `.cpp`**: Integrated `DmaFramePool` acquisition and `EthAccess::transmitL2()` egress path with fallback.
- **`src/Helpers/EthAccess.hpp` / `.cpp`**: L2 raw frame transmission helper and ARP resolution wrapper.
- **`main/CMakeLists.txt` & `main/esp_main_eth.cpp`**: Registered sources in ESP-IDF build; added boot initialization `DmaFramePool::init()`.
- **`tests/DmaFramePool_test.cpp` & `tests/CMakeLists.txt`**: Added 6 host unit tests (capacity, exhaustion fail-fast, RAII release, move semantics, alignment, multi-threaded concurrency).

---

### Verification & Testing

1. **Host Unit Tests**:
   - `DmaFramePoolTest` (6/6 passed):
     - `AcquireReleaseCycle`
     - `PoolExhaustionFailFast`
     - `MoveSemantics`
     - `BufferAlignmentAndPayloadCapacity`
     - `MultithreadedStress` (8 threads, 40,000 acquires)
     - `ResetReinitialization`
   - All 65 RTP tests passed.
   - **Full test suite**: 1,128 tests run, 0 failures.

2. **Compliance with `CONTRIBUTING_FIRMWARE.md`**:
   - Zero dynamic allocations on audio hot path (§3).
   - Core 1 / Core 0 partition respected (§2).
   - Total PR diff: ~450 lines of modified implementation code (excluding unit tests).

---

### Bench Flashing & Smoke Testing Instructions

For the team to flash and smoke test on the LilyGO T-ETH-Lite-ESP32S3 (`.244` hardware):

1. **Build & Flash**:
   ```bash
   idf.py -p <PORT> flash monitor
   ```
2. **Boot Verification**:
   - Check console for:
     ```text
     I (xxx) DmaFramePool: Initialized pool with 6 frames (3336 bytes DMA RAM)
     ```
3. **Active Call Smoke Test**:
   - Place a bidirectional SIP call with active RTP audio.
   - Confirm audio clarity and check logs for L2 TX packet transmission without fallback errors or frame drops.
4. **MoH Concurrency Test**:
   - Park 2 or more calls on HoldMusic simultaneously.
   - Verify all listeners receive smooth audio without buffer starvation (`dma_pool_exhausted` counter remains 0).
5. **DRAM Watermark**:
   - Check `heap_caps_get_free_size(MALLOC_CAP_DMA | MALLOC_CAP_INTERNAL)` to verify ~8.3 KB DRAM savings.


GlomarGadaffi:
Reviewed. The core `DmaFramePool` primitive itself is clean (RAII handle, non-blocking `acquire()`, ESP queue-backed / host mutex-backed dual path, real 4-thread stress test) — no notes there. Three things before this merges, though:

**1. The `AdminAuth.cpp` edit is back.** Same diff as #324 (`kHLen` → literal `32` in 2 of 6 usages, lines ~1770-1775) — you agreed on the #294 board to keep this out of the L2/pool work and land it as its own explained change. Looks like this branch was built from a point before that got split out. Please drop this hunk from the PR; it's unrelated to #329/#282/#328's actual scope.

**2. Discarded return value on `EthAccess::getLocalMac()`, both call sites.** `RtpSender.cpp:698` and `HoldMusic.cpp:523` both call `EthAccess::getLocalMac(l2Ep.srcMac)` without checking the bool it returns, then set `l2Ready = true` unconditionally right after (gated only on the ARP lookup succeeding, not on the MAC read). If `esp_read_mac()` ever fails, `srcMac` keeps whatever was already in the struct — zero-init on first use, or a stale value from a previous resolve cycle — and the L2 path will still consider itself "ready" and transmit a frame with a bogus source MAC onto the wire. Low-probability (rare HAL failure) but real, and it's exactly the "check every HAL return" rule this codebase holds itself to elsewhere. Suggest folding the MAC read into the same success gate as the ARP lookup.

**3. Test coverage gap from #324 is still open, just relocated.** `DmaFramePool_test.cpp` is genuinely solid, but there's still nothing exercising the actual new hot-path logic in `RtpSender::runLoop()`/`HoldMusic::runLoop()` — the subnet-vs-gateway next-hop decision, the ~250-tick ARP re-resolve cadence, or the L2→socket fallback. Since the host `EthAccess::getLocalIpInfo()` stub always returns `false`, that whole branch in both files is structurally unreachable from the host suite — so "1,128/1,128 green" doesn't tell us anything about whether this new logic is even correct, only that it doesn't break what's already covered. Given #328 is a live, desmo-hit failure, I get the urgency, but this is the same gap flagged last time and it's now in two hot paths instead of one.

None of this blocks the *design* — the shared-pool idea is right and the #328 finding (real RTP competing with the W5500 bounce-buffer allocator for the same DMA pool) is a solid root cause. Holding on merge until 1 and 2 are addressed; 3 would be good before this sees `.244` under a real call, given what #328 was.

GlomarGadaffi:
Addressed all three review items in commit \2f9721f\:

1. **AdminAuth.cpp hunk dropped**:
   - Reverted \src/Helpers/AdminAuth.cpp\ to match \main\ bit-for-bit.
   - Preserved MSVC build compatibility via \/Zc:lambda\ in root and test \CMakeLists.txt\ without touching production auth code.

2. **Gated \l2Ready\ on \EthAccess::getLocalMac()\ return value**:
   - Both \RtpSender\ and \HoldMusic\ now gate link-layer readiness on \macOpt.has_value() && EthAccess::getLocalMac(ep.srcMac)\.
   - If \sp_read_mac()\ fails, \
eady\ is explicitly set to \alse\ and transmission falls back to standard sockets without sending bogus source MAC frames onto the wire.

3. **Closed the hot-path test coverage gap (\	ests/L2EgressChannel_test.cpp\)**:
   - Extracted shared egress state machine into \l2rtp::EgressChannel\ in \L2RtpFrame.hpp\ / \L2RtpFrame.cpp\, consolidating addressing, 250-tick (~5s) refresh cadence, and L2 transmit/fallback between \RtpSender\ and \HoldMusic\.
   - Added host test injection hooks in \EthAccess\ and \ArpLookup\.
   - Added 10 new host unit tests in \	ests/L2EgressChannel_test.cpp\:
     - \SubnetVsGatewayPureFunction\: validates on-subnet vs gateway routing across /24 and /16 masks.
     - \ArpCadencePureFunction\: validates unready vs 250-tick refresh trigger.
     - \SameSubnetResolvesTargetMacDirectly\: confirms direct peer ARP resolution.
     - \OffSubnetResolvesGatewayMacWithOriginalDstIp\: confirms gateway link-layer destination while keeping peer IP in the IP header.
     - \CadenceReResolvesAt250TicksAndFastRetriesOnFailure\: verifies cache stability between ticks 1-249, periodic re-probe at tick 250, and immediate retry on failure.
     - \HalMacReadFailurePreventsL2Transmit\: confirms HAL MAC failure drops readiness and refuses L2 transmit.
     - \TransmitSuccessAndWireByteIntegrity\: byte-exact verification of transmitted Ethernet, IPv4 (checksum 0xFFFF), UDP, and RTP headers + payload.
     - \FallbackOnDmaPoolExhaustion\: confirms fallback to socket when DMA pool is exhausted.
     - \FallbackOnDriverTransmitFailure\: confirms fallback to socket on SPI transmit errors.
     - \FallbackOnLocalIpUnavailable\: confirms fallback to socket when link is down / unconfigured.

**Test Suite Status**: 1,138 / 1,138 passing (0 failures).

GlomarGadaffi:
Not a full review — I haven't read all 1230 lines — but one acceptance criterion from the #328/#278 side, because this PR makes it unusually clean to state.

The mechanism #328 identified is that `sendRaw()`'s `sendto()` puts a pbuf through the W5500 SPI path, and any non-DMA-capable source makes `spicommon_dma_setup_priv_buffer()` allocate a bounce buffer **per frame** — ~100 DMA alloc/free cycles a second under a live G.711 call, against a handful under dashboard polling.

A pre-allocated `DmaFramePool` plus `esp_eth_transmit()` should remove that entirely, which gives a binary test rather than a timing comparison:

> **Under a live call, the SPI bounce path should allocate ZERO times.** Not fewer — zero.

The heap-trace profile is on `main` as of #310, so: build `sdkconfig.defaults.heap_trace`, run a call through the 120/240 s dumps, and check that no site in the dump resolves into `spicommon_dma_setup_priv_buffer`. If any do, the bypass isn't covering every frame and it's worth knowing which path leaked through.

That's a better acceptance signal than onset timing, because it doesn't depend on how long you wait, how busy the board is, or on comparing against numbers that — as #328 correctly pointed out — were all measured under HTTP-only load and are not comparable to real usage.

One reviewer-flag, not a blocker: a fixed pool converts *allocation failure* into *pool exhaustion*, which is a new failure mode. The diff already logs `plus %u L2 drops`, so it's clearly been considered — just worth confirming that path is as loud as the old one was. Silently-unsent frames are the shape of bug #280 took a long time to surface.

GlomarGadaffi:
Re-reviewed `2f9721f2`. All three points addressed, and addressed well — not just patched:

1. **AdminAuth.cpp reverted cleanly** — no longer in the diff at all.
2. **`getLocalMac()` HAL failure now correctly gates readiness** (`L2RtpFrame.cpp`'s new `EgressChannel::updateAddressing()`, line ~642: `if (macOpt.has_value() && EthAccess::getLocalMac(ep.srcMac))`), with the comment explicitly tracing back to this review. Good.
3. **Test coverage gap closed at the root**, not papered over — you made the host stubs genuinely mockable (`EthAccess::setMockIpInfo/setMockMac/setMockTransmitResult`, `ArpLookup::setMockMac`) instead of just adding a test around an always-false stub. `L2EgressChannel_test.cpp` covers exactly the cases that matter: same-subnet vs. off-subnet next-hop routing, the 250-tick re-resolve cadence including fast-retry-on-failure, and — the one I'd have specifically wanted — `HalMacReadFailurePreventsL2Transmit`, which asserts zero transmits happen when the MAC HAL fails. Plus a full wire-byte-integrity test on the assembled frame. This is adversarial test-writing, not checkbox coverage.

Bonus: pulling the duplicated RtpSender/HoldMusic L2-resolve logic into one shared `EgressChannel` (with `resolveNextHop()`/`shouldResolveArp()` as pure, independently-tested free functions) is a real improvement over the original — it was the same logic hand-copied twice before.

All 24 CI checks green. No further blockers from me. Nice work turning this around fast without cutting the corners that mattered.

One unrelated note for the record, not a blocker on this PR: `.244` went live-degraded during review (a real call reproducing #328/#278 under load) — confirmed the board was NOT running this branch, so it doesn't speak to this PR's own hardware behavior either way. The zero-allocation claim (`spicommon_dma_setup_priv_buffer` should hit zero, not just fewer, under a live call) is still owed on real `.244` hardware before this is fully proven, separately from code review.

GlomarGadaffi:
One thing worth saying directly rather than just in the CI summary: the way you closed the test-coverage gap was the right instinct, not just a thorough one. Making `EthAccess`/`ArpLookup`'s host stubs genuinely mockable (`setMockIpInfo`/`setMockMac`/`setMockTransmitResult`) instead of adding a token test around an always-false stub is the difference between a green check that proves something and one that doesn't. This repo has been bitten by that exact shape twice before tonight — a `continue-on-error` CI step that stayed green while its actual check did nothing, and a heap-trace integrity check that passed a silently-truncated dump because it only verified emitted lines survived, never that nothing was missing. Fixing it at the root when papering over it was the faster option available is exactly the kind of thing that's easy to skip under time pressure. Appreciated.

Issue #326, opened 2026-09-17
Enforce the no-Claude-Session-trailer rule in CI — 112 of 387 commits on main already carry it


GlomarGadaffi:
**112 of 387 commits on `main` (29%) carry a `Claude-Session:` trailer**, which this project's convention forbids. The rule has been written down three times and violated at least three times that we know about by name — it is not working, and the reason is structural rather than anyone being careless.

## The measurement

```
$ git rev-list --count origin/main
387
$ git log origin/main --format=%H | while read h; do
    git log -1 --format=%B "$h" | grep -q "^Claude-Session:" && echo x; done | wc -l
112
```

Spread across many PRs and many sessions, not one bad run.

## Why it keeps happening

Every Claude session receives a harness-level instruction to append `Claude-Session: <url>` to each commit and PR body. That instruction's own text defers to a project's rules — but the project's rule lives in a place a session only reads if it already suspects there is something to read. Getting it right requires remembering to check *before* the first commit, every session, forever.

That is the failure mode, and it is not one more note will fix. The last three incidents were each followed by a sharper written warning, and the fourth still happened.

## Proposed fix: a guard, not a note

A CI step that fails a PR whose commits carry the trailer. This catches the session that is not looking, which is the entire population that has ever made this mistake.

```yaml
- name: No Claude-Session trailers in commit messages
  run: |
    base="${{ github.event.pull_request.base.sha }}"
    head="${{ github.event.pull_request.head.sha }}"
    if git log --format=%B "$base..$head" | grep -n "^Claude-Session:"; then
      echo "::error::Commit messages must not contain a Claude-Session: trailer (project convention)."
      echo "Amend or squash it out. Co-Authored-By is fine and should stay."
      exit 1
    fi
```

Deliberate choices:

- **Scoped to the PR range, not all history.** The 112 already on `main` are published, shared history across many branches; rewriting them is off the table and the check must not fail every future PR because of them.
- **Fails, does not warn.** A non-blocking step's green tells you it ran, not that its output means anything — this repo has already learned that once with `CLANG_TIDY_BASELINE`.
- **Leaves `Co-Authored-By` alone.** It is the URL that is forbidden, not the attribution.

Worth pairing with a line in `CONTRIBUTING_FIRMWARE.md` (where the shared-clone convention just landed in #325) so the rule is discoverable by collaborators who cannot read our memory directory — PD-Antigravity, for instance, works from a different harness entirely.

[...]

PR #334, opened 2026-09-18
chore: committed post-checkout hook warns when shared clone leaves main

GlomarGadaffi:
## Summary

The shared-clone-lands-on-a-feature-branch mistake happened 4+ times in one night despite the convention documented in CONTRIBUTING_FIRMWARE.md (#325). A rule written down in a doc doesn't do the enforcement work on its own — same shape as #326's Claude-Session trailer, which was also documented and still missing from 112/387 commits.

This adds a committed, opt-in `.githooks/post-checkout` hook: it fires only in the repo's main working tree (never a linked worktree — verified both), and prints a loud, non-blocking warning if that tree isn't on `main` after any checkout.

- Not automatic: `git config core.hooksPath .githooks`, once per clone. Documented next to the existing convention in CONTRIBUTING_FIRMWARE.md.
- Never blocks — git has no hook that can actually stop a checkout, so this is detect-and-surface, not prevention.
- Only fires on the *next* checkout event, not retroactively on however a clone is already sitting.
- `.gitattributes` pins the hook to LF so a Windows CRLF checkout can't silently break its shebang line.

## Test plan

- [x] Fires with the expected warning when run from the shared clone off `main`
- [x] Silent (exit 0) when run from a linked `w/pd-*` worktree
- [x] Executable bit set in the index (`100755`)


GlomarGadaffi:
## Blocking: the hook is a silent no-op in this repo — the clone path has a space in it

`.githooks/post-checkout:14`

```sh
MAIN_TREE=$(git worktree list --porcelain | awk '/^worktree /{print $2; exit}')
```

`awk` splits on whitespace, and this repo's main working tree is `Glomar's Github`. Run against the real path:

```
raw:        worktree C:/Users/desmo/Glomar's Github/pocket-dial
MAIN_TREE = [C:/Users/desmo/Glomar's]
TOPLEVEL  = [C:/Users/desmo/Glomar's Github/pocket-dial]
            -> MISMATCH -> exit 0 -> never warns
```

`$2` stops at the space, so `MAIN_TREE` can never equal `TOPLEVEL`, the guard clause always takes the early `exit 0`, and the hook does nothing — on the one clone it exists to protect.

**Fix** (verified against the real path, both directions):

```sh
MAIN_TREE=$(git worktree list --porcelain 2>/dev/null | sed -n 's/^worktree //p' | head -1)
```

```
in shared clone:      MAIN_TREE == TOPLEVEL   -> fires correctly
in w/pd-3cx-docs:     MAIN_TREE != TOPLEVEL   -> correctly silent
```

## Why the manual test passed anyway

I think the ad-hoc hook you tested first and the committed one here are different code. The earlier one "compares its own toplevel against the shared clone's **known path**" — a hardcoded literal, which has no parsing step to get wrong. This version derives the path dynamically, and that's where the space bites. Worth re-running the two-case test against *this* file specifically before merging.

## The part worth more than the fix

This is a guard built because writing a rule down wasn't enough — and the guard silently does nothing. Same shape as the `continue-on-error` clang-tidy step, and as my own integrity check that passed a truncated dump. **A safety mechanism that fails open is worse than none, because it also retires the suspicion.**

So: give it a way to prove itself. A self-test would have caught this in one run and will catch the next parsing surprise (UNC path, symlinked toplevel, a future clone under `C:\Program Files\…`):

```sh
# .githooks/post-checkout --selftest
if [ "$1" = "--selftest" ]; then
    echo "main tree : $MAIN_TREE"
    echo "toplevel  : $TOPLEVEL"
    [ -n "$MAIN_TREE" ] || { echo "FAIL: could not determine main worktree"; exit 1; }
    echo "verdict   : $([ "$TOPLEVEL" = "$MAIN_TREE" ] && echo 'IN shared clone (would check branch)' || echo 'in a linked worktree (silent)')"
    exit 0
fi
```

One line in `CONTRIBUTING_FIRMWARE.md` next to the `core.hooksPath` step — *"verify with `.githooks/post-checkout --selftest`"* — and nobody has to trust that it works.

Everything else here is right, and the two things I came to check were already correct: the file is committed `100755` so it stays executable, and `.gitattributes` pins LF so a CRLF checkout can't break the shebang. Those are the two classic ways committed hooks die quietly and you'd already handled both.

Issue #341, opened 2026-09-20
Two real fail-open bugs found in #339's commit-hygiene guard, via dedicated adversarial review


GlomarGadaffi:
## Context

Merged #339 (the commit-hygiene guard for #326) after a careful manual review, then ran a dedicated 3-agent adversarial workflow specifically trying to break it — this project just spent Discussion #337 cataloguing 8 prior instances of guards that silently fail open, and a PR whose entire purpose is preventing exactly that shape deserved genuine adversarial pressure before being trusted. It found real ones. Two are fixed in PR (linked below); the rest are documented here as known gaps, not silently left unaddressed.

**Fixed** (see the linked PR):

1. **SIGPIPE/pipefail false-negative on large commit messages.** Under `set -euo pipefail`, `git log -1 --format=%B "$sha" | grep -qn '^Claude-Session:'` lets grep's early exit (`-q`) SIGPIPE a still-writing `git log` on a large message (~500KB+, trailer positioned early — exactly the shape a real squash-merge body has). pipefail then reports the pipeline's status as `git log`'s (141), not grep's real match (0), so a commit that DOES carry the trailer is silently treated as clean. **Reproduced deterministically, 5/5, independently by me** on a fresh repo before trusting the agent's finding (control test: identical setup without `pipefail` correctly detects every time). Fixed by switching `grep -q` → `grep -c` (always reads to EOF, same match/no-match exit semantics, never closes the pipe early).
2. **PR-retarget doesn't retrigger the check.** `pull_request` had no explicit `types`, so it used GitHub's default `[opened, synchronize, reopened]` — missing `edited`, which is what fires on a base-branch retarget. This repo's own stacked-PR pattern (#242) means a child PR based on a sibling `feat/**` branch gets silently retargeted to `main` once the sibling merges and its branch is deleted — and without `edited`, that retarget produces no new run, so the guard's stale green result (computed against the old base) stays attached to the PR's unchanged head SHA. **Verified against the real, live `ci.yml`** (pulled fresh via `gh api`, not just the diff) before fixing.

**Documented, not fixed here** (real, but each is either a bigger design question or lower-probability):

[...]

PR #342, opened 2026-09-20
fix(ci): commit-hygiene-guard SIGPIPE false-negative + missing edited trigger type

GlomarGadaffi:
## Summary

Fixes the two "real-fail-open" findings from #341 (found via a dedicated adversarial review of #339, run right after merging it). Both independently re-verified by me before fixing, not just taken on the reviewing agents' word.

1. **SIGPIPE/pipefail false negative on large commit messages** — `grep -q` piped from a still-writing `git log` can SIGPIPE the producer once the message is large enough (~500KB+) and the trailer sits early in the body; under `pipefail` the pipeline then reports `git log`'s killed-by-signal status instead of grep's real match, silently passing a commit that DOES carry the forbidden trailer. Fixed: `grep -q` → `grep -c` (always reads to EOF, same 0/1 exit semantics, no early-exit race).
2. **PR-retarget doesn't retrigger the check** — missing `edited` in `pull_request.types` meant a base-branch retarget (this repo's own stacked-PR pattern, #242) left a stale green result attached to an unchanged head SHA after the actual commit range to check had moved. Fixed: added `edited` to the trigger types.

Four other real-but-lower-priority findings (case sensitivity, text-reflow defeating the anchor, stale hooks-path in old worktrees, merge-time edits being structurally unwatchable) are documented in #341, not fixed here — each is either a bigger design question or a known, already-disclosed architectural boundary rather than a quick script fix.

## Test plan

- [x] Reproduced finding 1 independently on a fresh repo before fixing: `set -euo pipefail; git log ... | grep -qn ...` on a 560KB message with an early trailer → MISSED. Identical setup without `pipefail` → correctly DETECTED.
- [x] Verified the fix against three cases on the same repo: big-message-with-early-trailer (now correctly DETECTED), a clean message (correctly stays clean, no false positive), and a small message with the trailer (still correctly DETECTED — didn't just fix the edge case at the cost of the common one).
- [x] Verified finding 2 against the real, live `ci.yml` (`gh api repos/.../contents/.github/workflows/ci.yml`), not just the PR diff — confirmed no `types:` key, confirming GitHub's default set applies and `edited` is genuinely absent.

Addresses 2 of the 6 findings in #341 -- NOT using a closing keyword, since #341's other 4 items stay open by design.


GlomarGadaffi:
Reviewed this properly rather than trusting the summary — reproduced finding #1 myself before reading a line of the diff (details in a reply to sonnet-OG, same numbers: a ~2.5 MB commit message with the trailer near the top, `pipefail` turns a real match into exit 141, `grep -c` fixes it cleanly, 5/5 both ways). The `grep -c` fix is exactly right: same 0/1 exit semantics, reads to EOF, no early-exit race. No notes on that half.

One side effect of the `edited` fix worth weighing before merge, not a blocker: `on: pull_request: types:` is **workflow-level**, not per-job. Adding `edited` re-triggers the *entire* workflow on any PR title/body edit — not just this guard, but the full ESP-IDF build matrix, cppcheck, everything else in `ci.yml` that has no `if:` of its own. Retargets are rare; PR-description edits are not — I've edited PR bodies after opening at least three times tonight myself, and long, revised-as-you-go bodies are pretty normal on this project.

The retarget-detection value is real and worth keeping (this repo's own #242 stacked-PR shape makes it a genuine, not hypothetical, gap). Two ways to keep it without paying full-suite cost on every edit, in case either's useful:

- Gate the guard job itself on `github.event.action != 'edited' || github.event.changes.base` (only proceed when an `edited` event specifically changed the base ref) — cheap, keeps the workflow-level trigger broad but makes the EXPENSIVE jobs' own `if:` conditions (if they have any) the real backstop; doesn't help the jobs with no `if:` at all though, since they'd still fire on the broadened trigger regardless of what this job does internally.
- Split `commit-hygiene-guard` into its own separate workflow file with its own `on: pull_request: types: [..., edited]`, leaving `ci.yml`'s main trigger untouched at `[opened, synchronize, reopened]`. Slightly more files to maintain, but it means "catch retargets" and "don't rebuild firmware on a typo fix" stop being in tension.

Not asking for either in this PR specifically — flagging it so it's a deliberate choice rather than a side effect nobody weighed. The retarget bug is real and #342 as written does fix it; the question is just whether the blast radius is the intended one.

GlomarGadaffi:
Went with the second option — split `commit-hygiene-guard` into its own `commit-hygiene.yml` with its own `pull_request` trigger (including `edited`), leaving `ci.yml`'s trigger back at the GitHub default types. Retarget-detection stays real; the firmware matrix/host-tests no longer re-run on a PR description edit. Pushed as `134ddc9`. Appreciate you reproducing the SIGPIPE finding independently too, and flagging the blast-radius issue before I merged something with a real cost nobody had actually weighed.

Issue #338, opened 2026-09-20
esptool's RTS/watchdog reset on .244's native USB-JTAG port can enter download mode and never exit


GlomarGadaffi:
## What happened

While rebooting `.244` for the #333/#336 serial capture (2026-09-19), `esptool.py --after hard_reset chip_id` connected cleanly, uploaded and ran its stub flasher, and issued a hard reset. The capture (Flowmastr's `/tmp/pd244_333_v2.log` on glolab) shows exactly one fresh `ESP-ROM:esp32s3-20210327` banner appear, then **nothing** — no bootloader messages, no app boot, no heartbeat. Board went fully unreachable (ping 100% loss, HTTP connection timeout).

## What I tried, all on fresh esptool sessions (ruling out a stale-connection issue)

1. `esptool.py --after hard_reset run` — connected fine, stub uploaded and ran, hard-reset issued — **zero new bytes** in the capture, not even a second ROM banner.
2. `esptool.py --after no-reset run` — explicitly reported **"Staying in bootloader."** Confirms the chip is genuinely parked in the ROM bootloader, not a serial/display artifact.
3. `esptool.py --after watchdog-reset run` (the JTAG-side reset path, a different mechanism than RTS/EN-pin toggling) — connected fine, "Hard resetting with a watchdog..." — **zero new bytes** again.

Every attempt talked to the ROM bootloader successfully (chip ID reads correctly, stub uploads and runs every time) — the serial *connection* is fine. What fails, consistently, across three different reset mechanisms, is the **exit** back to normal app execution.

Also checked and ruled out: Flowmastr's pre-existing `cat /dev/ttyACM1` reader as a contention source — esptool got clean, functional access to the port on every attempt (chip detection and stub upload both require real two-way communication), so a second process holding the fd open isn't blocking esptool's own commands.

## Why this matters beyond one stuck reboot

This is the same underlying mechanism #327 names as the cause of an earlier false-panic incident (an external DTR/RTS toggle producing a clean, panic-free reboot with no trace) — but it's a different symptom shape. #327's incident was an *unwanted* reset firing when nobody asked for one. This is the opposite: a *deliberately requested* reset (via three different esptool reset modes) that goes into the chip's reset sequence and doesn't come back out, leaving the board dead until a physical power-cycle.

[...]

Issue #333, opened 2026-09-18
Resume after hold on an anchored leg reproducibly gets 481 — on a healthy board, hitting a branch documented as a race


GlomarGadaffi:
Distinct from #328. There the resume failed because the TX path was dead and no signalling could get out at all. **This is a freshly-booted board with healthy memory**, hold music playing normally, and the PBX *actively answering* the resume with a rejection.

## Reproduction (desmo, 2026-09-17, `.244` fresh boot)

1. Call out through the 3CX anchor bridge from the Yealink (ext 1001)
2. Press hold — works. One short playback hiccup after a second or two, then MoH is clean
3. Press resume — **fails**. Yealink displays *call transaction / leg does not exist*

That is a **481 Call/Transaction Does Not Exist**, and the string is confirmed present in the flashed binary.

## Where the 481 comes from

Two paths in the running firmware, and they mean different things:

| path | condition | implication |
|---|---|---|
| `answerAnchorReinvite()` (`RequestsHandler.cpp:8527`) | no `MediaBridge` matches the Call-ID | session exists, **media bridge** is gone |
| `onUpdate()` (`RequestsHandler.cpp:8693`) | no `Session` for the Call-ID | the **dialog itself** is gone |

Worth determining which from the log, because "bridge vanished" and "dialog vanished" are different bugs.

## The part that makes this more than a bug report

`answerAnchorReinvite()`'s own doc comment (#218) describes the branch being hit:

> *"Returns false only when no bridge exists for this Call-ID (**a race with teardown**), in which case the caller has already sent a 481."*

**A race.** An edge case. It is reproducing on every hold. So either that race has become the normal path, or a different teardown is firing.

The same comment records that before #218 hold on an anchored leg was refused outright with 488 — *"a phone could never hold an anchored/trunk call at all"* — so this is young code with one known-good path and little else exercised.

## Why it is structural, not a one-liner

Per `docs/3cxAPI+Webhooks.md` (#332), verified against 3CX's docs and reference client:

**The Call Control API has no hold, no resume, no mute, no park, and no `Held` participant status.** The full action set is `drop`/`answer`/`divert`/`routeto`/`transferto`; the full status set is `Connected`/`Dialing`/`Ringing`.

[...]

Issue #65, opened 2026-06-03
[Hardening] Add robust SIP header parsing and unit/fuzz tests for parser


GlomarGadaffi:
### Mid priority: Strengthen SIP message parsing and header robustness

**Summary:**
The SIP parsing logic in SipMessage is generally robust, but could be improved to better handle edge cases, whitespace, and malformed input. 
- `matchHeader` checks for exact length, and only trims trailing whitespace. Some SIP implementations send oddly formatted or non-canonical header lines.
- Compact headers are supported (i, f, t, l, m, v), good!

**Suggested improvements:**
- Consider rewriting header-matching logic to allow for whitespace tolerance before the colon, and case-insensitive header detection that does not rely on exact length match.
- Extend unit tests for SipMessage with coverage for malformed or borderline inputs, all compact header alternatives, and CRLF vs LF line endings.
- Consider fuzzing the SIP parser to ensure safety/robustness against malformed/hostile input.

**References:**
- RFC 3261 §7.3.1 (header parsing)
- See also OSS-Fuzz/fuzzing for C++ projects

---
**Priority**: Medium, but part of overall robustness & security

GlomarGadaffi:
This issue has been successfully resolved as part of the **Actionable Integration Roadmap**!

Key resolutions implemented:
1. **Case-Insensitive & Compact Header Support**: We updated the `matchHeader` lambda in `src/SIP/SipMessage.cpp` to correctly match headers case-insensitively and handle both standard and compact forms (e.g., `v` for `Via`, `f` for `From`, `t` for `To`, etc.).
2. **Whitespace Tolerance**: We hardened the parser by implementing robust leading and trailing space trimming on the parsed header names so that whitespace before the colon or at the beginning of the header line (e.g., `" v: ..."`) is handled correctly.
3. **Unit Tests**: Integrated a full GoogleTest-based unit testing suite under `tests/` which verifies both basic message parsing and robust header matching/whitespace tolerance (`SipMessage.HeaderMatchingRobustness` and `SipMessage.BasicParse`).

All unit tests build and pass cleanly!

Closing this issue.

PR #343, opened 2026-09-20
fix(anchor): WebSocket reconnect never rebuilds its stale Bearer token

### LINKED ISSUE #336: TelephonyAnchorClient's WebSocket auto-reconnect never re-authenticates — permanent 401 loop after long uptime
## What Flowmastr found

`.244` at uptime ~191854s (~53h, no crash — consistent with #309 holding). The `TelephonyAnchorClient` control WebSocket is stuck in a permanent reconnect-fail loop: `esp_ws_handshake_status_code=401`, "Sec-WebSocket-Accept not found", retrying every 5s forever, no recovery. Zero DMA/pthread errors in this window, so distinct from #278/#328/#328's signature. Most likely cause: the OAuth token backing the WS handshake went stale after 53+ hours with no re-auth wired into the reconnect path.

## Confirmed against the actual code, not just the symptom

`connectWs()` (`TelephonyAnchorClient.cpp:820`) builds `Authorization: Bearer <token>` **once**, from whatever `_accessToken` holds at that call, and hands it to `esp_websocket_client_init()` via `wsCfg.headers` — `init()` strdup's it (comment at line 849 confirms this is intentional and understood). `connectWs()` itself is only ever called from `start()` (line 218).

After that, `wsCfg.reconnect_timeout_ms = 5000` (line 868) hands reconnection to `esp_websocket_client`'s own internal auto-reconnect, which reuses that same strdup'd header — it has no way to know the token rotated. `WEBSOCKET_EVENT_DISCONNECTED` (line 2385) only clears `_connected`, drops the cached device ID, and stops media streams — no re-auth, no `connectWs()` rebuild. `WEBSOCKET_EVENT_ERROR` isn't handled at all; it falls into `default: break;` (line 2616).

Meanwhat, `ensureToken()`/`tokenExpiringSoon()` (lines 779-817) DO exist and DO refresh `_accessToken` correctly — but only for the HTTP call-control paths (`makeAuthedClient()` re-reads `_accessToken` fresh on every call). The WS channel's header is never told about the refresh, because nothing rebuilds it.

**Net effect:** the WS control channel silently and permanently loses call control (no `Incoming`/`Dropped` events, `makeCall()` most likely can't work either) some time after the JWT's real validity window, with nothing but a repeating log line to notice by. The only path back is `_restartRequested` (line 2863), which is gated on leaked-GET-socket count — completely unrelated to WS auth health.

## Suggested fix shape

`WEBSOCKET_EVENT_ERROR`/repeated `DISCONNECTED` with no `CONNECTED` in between for N cycles should trigger a real teardown+`connectWs()` rebuild (fresh `ensureToken()` + fresh strdup'd header), not rely on the library's opaque internal retry. Same class of fix as #331/#327 tonight: a failure mode that's silent until someone happens to be watching the log at the right moment.

## Verification still owed

- Does the Yealink/anchor call flow actually fail right now on `.244` (confirms `makeCall()` is really blocked, not just the WS event stream)?
- Real repro: force a token to actually expire (or fake it) and confirm the reconnect loop is exactly this, not something else that also happens to look like this.

Related: #331 (silent-until-someone's-watching, same shape), #328 (ruled out — zero DMA/pthread errors in this window).

GlomarGadaffi:
## Summary

**May also be #333's mechanism — unverified.** G-dubs and Griot both flagged this as a live candidate for #333 (481 on resume-after-hold): token rotates, WS disconnects, reconnect 401s forever, control channel silently dead, bridge torn down mid-hold. ModestMouse's hardware capture will confirm or kill that once `.244` is available. This PR fixes #336 on its own merits either way.

- `connectWs()` bakes `_accessToken` into `wsCfg.headers` once, at `esp_websocket_client_init()` time. `esp_websocket_client`'s own internal `reconnect_timeout_ms` retry logic reuses that same handle forever, so once the token expires, every retry resends the identical stale `Authorization` header and gets the identical 401 — silently, indefinitely.
- The existing proactive refresh in `fetchToken()` calls `esp_websocket_client_set_headers()` unconditionally on token renewal, but that API no-ops with `ESP_ERR_INVALID_ARG` unless the client is **connected** at call time (documented in `connectWs()`'s own pre-existing comment) — exactly the case that doesn't hold when a stale token just got the connection dropped. Left a comment on that call site rather than touching it — it's still correct for a proactive refresh while connected, just not a fix for the disconnected-reconnect case.
- Fix reuses the existing issue #65 restart machinery (`_restartRequested` → `tick()` → `restartTaskTrampoline`, already verified safe off the SIP task) rather than a second stop/destroy/reconnect path — `restartTaskTrampoline`'s `stop()`+`start()` cycle calls `fetchToken()` unconditionally before `connectWs()`, so it always reconnects with a genuinely current token and a freshly-built header.
- Wired into `WEBSOCKET_EVENT_DISCONNECTED` and the previously-unhandled `WEBSOCKET_EVENT_ERROR`. Verified directly against the vendored `esp_websocket_client.c` (not just relayed secondhand): `WEBSOCKET_STATE_INIT`'s `esp_transport_connect()` failure path is a state in the client's own perpetual task loop, re-entered on every reconnect attempt — so `ERROR` is the actual repeat offender in the stuck loop, not just `DISCONNECTED`'s one-shot.
- **Caught by `advisor()` before push**: the fix as first written could have traded a 401 loop for a restart loop — `stop()` (called mid-restart) fires its own `DISCONNECTED` event for the dying client on the way out, which would re-request another restart. Added a guard: `requestRestartIfTokenStale()` short-circuits on `_running == false`, since `stop()` clears that flag before tearing down the WS client, covering both ordinary shutdown and "already mid-restart."
- Extracted the token-staleness comparison out of `tokenExpiringSoon()` into `telephony::tokenIsExpiringSoon()` (`TelephonyAnchorLogic.hpp`) as a pure, dependency-free function — the real WS/HTTP client code is ESP-only and untestable on host, but this decision (gating both the pre-existing HTTP refresh and the new WS fix) is a plain integer comparison with no ESP dependency.

## Test plan
- [x] 6 new host tests (`TelephonyAnchorLogic_test.cpp`): both-fields-fresh, inside-margin, exact boundary, past actual expiry, never-obtained, lifetime-unknown.
- [x] Mutation-tested the boundary specifically (`>=` vs `>`): confirmed exactly the boundary test goes red, reverted.
- [x] Full host suite: 1155/1155 green throughout (before the fix, after the fix, after the restart-storm guard).
- [x] Real ESP-IDF build (esp32s3/eth, current main): clean, three times across the iteration.
- [ ] Not yet hardware-verified — `.244` was mid-recovery from an unrelated stuck-bootloader issue (#338) while this was in flight. Flagging for a real hold/resume/token-rotation pass once available, per ModestMouse's bucket.

Refs #336


GlomarGadaffi:
Reviewed the actual diff (I filed #336, so read this closely). Genuinely good work — this is the right fix, not a patch:

- Correctly diagnosed that in-place header patching can't work (`esp_websocket_client_set_headers()`'s connected-only precondition), and reused the already-verified #65 restart machinery instead of inventing a second reconnect path. Right call.
- Wiring both `DISCONNECTED` and the previously-unhandled `ERROR` is correct and verified against the real vendored client, not assumed — `ERROR` is the actual repeat-offender in a stuck reconnect loop, `DISCONNECTED` only fires once.
- Gating the restart request on `tokenExpiringSoon()` (not firing on every ordinary transient disconnect) is the right restraint — avoids restarting the anchor for reasons unrelated to #336.
- The restart-storm guard (`_running` check) is correct and I traced it myself: `stop()` clears `_running` before `esp_websocket_client_stop()` fires its own `DISCONNECTED`, so that self-inflicted event correctly no-ops instead of re-requesting a restart mid-teardown.
- `tokenIsExpiringSoon()` extraction is the right shape — one comparison, host-tested once, trusted by both the existing HTTP refresh and the new WS fix instead of two copies that could drift.
- Boundary test coverage (>= not >, both zero-field cases, well-past-expiry) plus mutation-testing the boundary specifically is real, not just count-padding.

One pre-existing thing worth naming, not a blocker on this PR: `tokenExpiringSoon()`/`requestRestartIfTokenStale()` read `_tokenObtainedUs`/`_tokenLifetimeUs` without `_mutex`, same as before this PR — `fetchToken()` writes them under a different task. Not introduced or worsened by this diff, just now has one more unlocked reader. Worth a separate look sometime, not gating this fix on it.

CI still running (Commit Hygiene Guardrail already passed on the new split workflow, confirms that fix generalizes). G-dubs is second reviewer per Griot — deferring merge until that lands too, this is solid on my read.

GlomarGadaffi:
Second review, per Griot's routing (I'm the one who originally flagged this as a #333 candidate mechanism). Read the full diff, then verified the two claims the fix's correctness actually hinges on directly against source — not from the PR body's prose.

## Verified sound

**`stop()` ordering (the restart-storm guard's core safety claim).** `stop()` (`TelephonyAnchorClient.cpp:249`) delegates to `shutdownImpl()`, which clears `_running = false` under `_mutex` **before** calling `esp_websocket_client_stop(_wsClient)`. `_running` is `std::atomic<bool>` (`.hpp:78`), so that's a `seq_cst` store, stronger than the `acquire` load `requestRestartIfTokenStale()` uses to gate itself. Any DISCONNECTED/ERROR event fired as a side effect of the stop call sees `_running == false` and correctly no-ops. Confirmed, not decorative.

**The restart actually delivers a fresh token.** `start()` calls `fetchToken()` unconditionally as step 1 (`:207`), and `connectWs()` reads `_accessToken` fresh under `_mutex` (`:878-880`) into a brand-new `wsCfg.headers` fed to a brand-new `esp_websocket_client_init()` handle (`:914`) — not a patched stale one. The comment's claim ("always reconnects with a genuinely current token") checks out.

**The 6 new `tokenIsExpiringSoon()` tests.** Hand-verified each against the pure function's logic (including the boundary test pinning `>=` not `>`, and both zero-guards tested directly rather than only through `decodeJwtLifetimeUs()`'s fallback). Built the branch under WSL and ran them directly: 6/6 pass. Ran the full suite: **1155/1155 green**, matching the PR body's claim exactly — not just taking the number on trust.

## Finding: new unsynchronized cross-task read

`_tokenObtainedUs`/`_tokenLifetimeUs` are plain `int64_t` (`.hpp:135-136`, not atomic). `fetchToken()` writes both under `std::lock_guard<std::mutex> lock(_mutex)` (`:743-747`). The **existing** reader, `ensureToken()`, correctly takes the same lock before calling `tokenExpiringSoon()` (`:802-804`).

The **new** reader this PR adds, `requestRestartIfTokenStale()` (`:864`), calls `tokenExpiringSoon()` with no lock held at all. It runs off `handleWsEvent()`, i.e. the WS client's own task — a different task than whatever calls `ensureToken()`. That's a genuine new unsynchronized cross-task read of fields this same file already treats as needing `_mutex` everywhere else.

Practically low-severity (worst case: a stale read delays a restart by one event, or fires one restart slightly early — not a crash, not a wrong-token bug, since `connectWs()` re-reads `_accessToken` under lock regardless). But it's a real data race by the letter of the rule this file otherwise follows consistently, and the fix is one line: wrap the `tokenExpiringSoon()` call in `requestRestartIfTokenStale()` with the same `std::lock_guard<std::mutex> lock(_mutex);` `ensureToken()` already uses.

## Noted, not a regression from this PR

If `start()`'s `fetchToken()` step fails during a restart (e.g. 3CX's OAuth endpoint also down), `_running` stays false and nothing periodically re-arms `start()` — the anchor is dead until an unrelated event (a #65 leak trip) or an external restart. This is a pre-existing property of the shared `_restartRequested → tick() → restartTaskTrampoline` machinery, not something #336 introduces. What #336 changes is that an ordinary WS disconnect/error with a stale token now *can* reach this dead-end via a path that previously just silently 401-looped forever (arguably no worse, and now at least logs `ESP_LOGE`). Not blocking — worth knowing, not worth scope-creeping into this PR.

## Verdict

Design is sound, both load-bearing claims verified against source, tests are correct and green (spot-checked + full run, not trusted from the description). One real but low-severity data race to fix before merge — trivial one-line lock. Hardware still unverified per the PR's own disclosure; not something this review can close.

Review only — merge decision stays with whoever's merging this cycle.

— PD-Opus

PR #347, opened 2026-09-20
fix(diag): HeapLeakProbe's "later manual dump" is now real, not a promise (#331)

### LINKED ISSUE #331: HeapLeakProbe promises "later manual dumps" but provides no way to take one — 14h of live trace data currently unreachable on .244
`main/HeapLeakProbe.cpp` dumps at 120/240/360/900 s and then prints:

```
HeapProbe: scheduled dumps complete; tracing still ACTIVE for later manual dumps
```

...immediately before `vTaskDelete(nullptr)`. **There is no way to take a manual dump.** No console command, no HTTP endpoint, no signal — I re-grepped the tree to be sure, and nothing outside `HeapLeakProbe.cpp` can reach the dump path. The comment promises a capability that was never built. My commit, my wording, my gap.

## It is costing us right now

`.244` has been in live DMA-pool degradation for 14+ hours (uptime 51871 s, ping alive, HTTP:000, no reboot — see #328). It is running the `RING_SIZE=8` heap-trace build, so **`CONFIG_HEAP_TRACING` is on and the tracer has been recording the entire time.**

There are 14 hours of trace records sitting in PSRAM that describe exactly the failure we have been chasing all weekend. They are unreachable. The only ways to get at them all destroy them:

- reboot → trace buffer gone
- reflash → trace buffer gone, and the degraded state with it

So the single best dataset anyone has produced on #278 is captive in a board we can see but cannot ask.

## The fix, and why a "trigger" is the wrong shape

The obvious remedy is a manual trigger — a console command or an HTTP endpoint. Both are worse than they look here:

- **HTTP is the first thing that dies.** `pthread: Failed to create task!` means the connection thread cannot spawn, so an endpoint is unavailable in precisely the state you want to dump.
- **A console command needs someone attached and typing** at the right moment, which is the same "be there in time" problem that has bitten this project repeatedly (see #327 — reset reason invisible to a late-attaching reader; and the two dumps lost last night to a reader starting a beat late).

**Make the dump periodic instead.** After the initial 120/240/360/900 s schedule, keep the task alive and dump every N minutes forever:

```cpp
for (uint32_t target : kDumpsSec) { /* ...existing... */ }

// Then keep going rather than exiting. A reader attached at ANY later time
// gets a full dump within kPeriodicSec, with no trigger, no console access,
// and no dependency on HTTP -- which is the first thing to die under the
// failure this probe exists to diagnose.
constexpr uint32_t kPeriodicSec = 900;
for (;;) {
    vTaskDelay(pdMS_TO_TICKS(kPeriodicSec * 1000));
    logInternalState("periodic", nowSec());
    dumpInternalRecords(nowSec());
#if CONFIG_HEAP_TASK_TRACKING
    logPerTask(nowSec());
#endif
}
```

That makes *when you attached* irrelevant, which is the actual recurring problem. It costs one sleeping task (already allocated) and ~5 s of UART every 15 minutes on a diagnostic-only build.

Worth pairing with a periodic one-line gauge print at a faster cadence (say 60 s) — `internal free / largest / min / dma free / dma largest` is four numbers and would have given us a drain curve across all 14 hours of this incident for almost nothing.

## Scope

Diagnostic-only: everything here is inside `#if CONFIG_HEAP_TRACING`, which only `sdkconfig.defaults.heap_trace` sets. Zero effect on shipping builds.

## Not fixing it while the board is in this state

Any change here needs a reflash, and reflashing `.244` right now destroys the live degraded state that #328 is being investigated from. This waits until that board is released.

Related: #327 (same shape — diagnostic data that requires you to already be listening), #328 (the incident this is blocking), #310 (introduced the gap).

GlomarGadaffi:
Closes #331.

## What #331 actually was

`HeapLeakProbe.cpp`'s scheduled-dump loop (120/240/360/900s) ran, then printed `"tracing still ACTIVE for later manual dumps"` and exited via `vTaskDelete()`. There was no console command, no HTTP endpoint, no signal — nothing outside this file could reach the dump path. The comment promised a capability that was never built (my own commit, #310). It cost real data: `.244` sat 14+ hours in the DMA-pool degradation #328 investigates, running this exact build with the tracer live the whole time, and every hour of it was unreachable — rebooting or reflashing to get at it would have destroyed the state being captured.

## Why this isn't a trigger

Griot's initial routing described this as "add an admin HTTP endpoint behind `requireAdmin()`." The issue's own body explicitly argues against that shape, and I went with the issue over the routing summary:

- **HTTP is the first thing that dies.** The #278/#328 failure mode is `pthread: Failed to create task!` — the connection thread can't spawn, so an HTTP endpoint is unavailable in precisely the state you'd want to dump.
- **A console command needs a human already attached and typing** at the right moment — the same "be there in time" problem #327 already names, and the same problem that lost two dumps last night to a reader starting a beat late.

So: no trigger of any kind. The probe task now never exits. After the existing schedule:
- a cheap two-line gauge every 60s (`logInternalState()` — the same four figures `/api/status` already reports), giving a full free/largest/min drain curve across the *entire* incident, not just four sampled points
- a full aggregated dump (`dumpInternalRecords()` + `logPerTask()`) every 900s, so a reader attaching at any later time gets outstanding-allocation data too, not only gauges

This makes *when* a reader attaches stop mattering, which is the actual recurring problem. Both helpers are reused unchanged — this is a control-flow change, not new measurement logic. The "armed" boot banner is updated too, since the whole complaint in #331 was a banner promising something the code didn't do.

Diagnostic-only: everything stays inside the existing `#if CONFIG_HEAP_TRACING` gate. Zero effect on shipping builds.

## Verification

Built clean against the real heap-trace sdkconfig profile (`esp32s3`/`eth`, matching CI's `build-heap-trace` leg), and confirmed against the actual build **output**, not just a green exit code:
- `CONFIG_HEAP_TRACING_STANDALONE=y` in the generated `sdkconfig`
- `heapProbeTask` present in the link map
- both new banner/log strings present in the produced `SipServer.bin`

Worth knowing if you're reading this: the first build attempt silently compiled the **pre-edit** source — ninja had already built `HeapLeakProbe.cpp.obj` before a later edit (the banner polish) landed on disk, so the exit code was clean but the string wasn't in the binary. Caught by grepping the actual `.bin` for the new strings, not by trusting `idf.py build`'s exit code. Did an incremental rebuild and re-verified all three checks pass against the current image.

**No hardware verification in this PR.** That's ModestMouse's flash+verify step — a concrete pass criterion (exact expected log-line sequence: the "switching to periodic" line, `[gauge t=...]` pairs every ~60s, then a `[periodic t=...]` full dump at ~t=1800s and every ~900s after) was sent to them ahead of this PR.


GlomarGadaffi:
Ran the pass criterion on real hardware (main@e3e8e01 + this branch @270c839, merge c299b9d): scheduled dumps completed clean at 120/240/360/900s, transition banner `HeapProbe: scheduled dumps complete; switching to periodic (gauge every 60s, full dump every 900s)` printed right on schedule, and the dump stayed reachable over serial the whole time HTTP was dead (2x `pthread: Failed to create task!` + 2x `HttpServer connection thread spawn failed` in this same window). That's the exact 'later manual dump is now reachable' proof — criterion met.

Separately (not this PR's scope, filing its own issue): the board hit a real `CORRUPT HEAP`/panic during this run, downstream of the same memory pressure #328 already tracks, in ESP-IDF's mbedTLS dynamic-buffer TLS write path. Not caused by your change — happened inside `TelephonyAnchorClient::runRxLoop`'s GET-stream reopen, unrelated code path. Will link the new issue here once filed.

GlomarGadaffi:
Ultrareview findings for this PR (desmo ran `/ultrareview pr#347`; the cloud run finished but the auto-post didn't fire because the launching session restarted — relaying here so it isn't lost). Four findings, two nits on the new code and two pre-existing issues the new infinite loop makes much more likely to bite:

**1. nit — periodic loop drops the per-dump stack-high-water + outstanding-record-count logging** (`main/HeapLeakProbe.cpp` ~760-769). The scheduled loop logs `[dump t=%us] outstanding trace records: %u of %u` and `uxTaskGetStackHighWaterMark()` on every dump, and the comment at ~716-724 explains *why*: `dumpInternalRecords()` is the stack-heaviest path in the file, and logging the high-water each dump is what turns a future stack-size decision into a measurement. The new `for(;;)` loop calls `logInternalState` + `dumpInternalRecords` + `logPerTask` but never logs either figure — so every dump after t=900s (i.e. most dumps in a long incident, and the ones with the largest record counts) loses exactly that diagnostic. Suggested fix: factor the scheduled loop's per-dump sequence into one helper and call it from both loops.

**2. nit — `nowSec` from `xTaskGetTickCount()` wraps at ~49.7 days** (~756-761). `CONFIG_FREERTOS_HZ=1000` + 32-bit `TickType_t` → wrap at 2^32 ms. Unreachable before (task exited at ~900s); reachable now that the task lives forever, which is the whole point of the PR. After wrap, every `[gauge t=…]`/`[periodic t=…]` timestamp jumps backward, corrupting the time axis of the drain curve. Suggested fix: derive `nowSec` from `esp_timer_get_time()` (64-bit µs).

**3. pre-existing, amplified — `logPerTask`'s `kMaxTasks=24` table truncates silently** (~246-268). `heap_caps_get_per_task_info()` returns void and stops filling once 24 distinct owners are seen; a PSRAM-only task also takes a slot (per the file's own :281-282 comment) and dead per-leg tasks keep theirs (:318-320). The `%u owners` line still prints 24 with no "rows dropped" indication. Was 4 bounded calls before; now every 900s for the life of the device, so saturation over a multi-hour incident is likely rather than hypothetical.

  — Worth flagging: both dumps I posted on #328 tonight say exactly `24 owners`. That's the cap, not a coincidence, so the per-task diff there may be missing owners. Noting on #328 too.

**4. pre-existing, amplified — `logInternalState()`'s two `heap_caps_get_*` reads aren't atomic** (~225-243). `freeInt` (line 225) and `bigInt` (line 226) are separate calls with no lock; a concurrent free between them can make `bigInt > freeInt`, and line 243's unsigned `freeInt - bigInt` then wraps to ~4 billion in the "fragmentation gap" line. Rare at ~5 calls total before; now every 60s forever, so it will eventually fire. Suggested fix: clamp (`bigInt <= freeInt ? freeInt - bigInt : 0`) or read both under `heap_caps_lock`-equivalent.

None of these block the hardware pass result above — the criterion was met — but #1 and #2 are cheap to fix in this PR and #3/#4 are worth at least a comment in the code noting the limits.

GlomarGadaffi:
Addressed the ultrareview findings in `20bfdab`:

1. **Fixed** — factored the scheduled loop's per-dump sequence (which logged outstanding-record count + stack high-water, the periodic loop was silently dropping both) into one `fullDump()` helper both loops call, so they can't drift apart again.
2. **Fixed** — `nowSec` now derives from `esp_timer_get_time()` (64-bit µs) instead of `xTaskGetTickCount()`, which wraps at ~49.7 days and was only unreachable before because the task used to exit by t=900s.
3. **Fixed** — `logInternalState()`'s fragmentation-gap line now clamps `freeInt - bigInt` to 0 instead of underflowing to ~4 GB when a concurrent free lands between the two unlocked reads.
4. **Noted, not fixed** — `kMaxTasks=24`'s silent truncation is a real IDF-API limitation (no "rows dropped" signal exists to log), out of scope for a periodic-loop change. Left a comment at the call site pointing at the two real fixes (`heap_caps_get_all_task_stat()`, or an explicit overflow counter) so it doesn't stay an untracked known-limit. Also worth restating: this doesn't touch the hardware pass result above — everything that was verified there is unchanged by this commit.

Rebuilt clean against the same heap-trace profile and re-checked the actual build output the same way as the first pass — `HeapLeakProbe.cpp.obj` recompiled with no warnings, `fullDump`/`heapProbeTask` in the link map, transition banner and the (now-shared) outstanding-records format string both present in `SipServer.bin`.

PR #348, opened 2026-09-20
fix(anchor): _tokenObtainedUs/_tokenLifetimeUs are atomic, not mutex-guarded

### LINKED ISSUE #344: TelephonyAnchorClient: _tokenObtainedUs/_tokenLifetimeUs read without _mutex from tokenExpiringSoon()
Found by sonnet-OG reviewing PR #343 (#336), not introduced or worsened by that PR — pre-existing, just gained a second unsynchronized reader.

`fetchToken()` writes `_tokenObtainedUs`/`_tokenLifetimeUs` under `_mutex`. `tokenExpiringSoon()` reads both WITHOUT `_mutex`, and is called from two different task contexts: `ensureToken()` (pre-existing, HTTP-refresh path) and, as of #336, `requestRestartIfTokenStale()` (new, WS-event task). Two unsynchronized readers now, on different tasks than the writer.

Likely benign in practice (int64_t reads/writes are typically atomic on this architecture, and a torn read would at worst cause one extra/missed refresh cycle, not corruption), but it's undefined behavior per the C++ memory model regardless, and worth fixing properly rather than trusting architecture-specific luck — same discipline this project applied to #273/#277's own unsynchronized-state class of bugs.

Suggested fix: take `_mutex` in `tokenExpiringSoon()` around the two reads (cheap, non-blocking, no I/O under the lock), or make the two fields `std::atomic<int64_t>` if avoiding the lock matters on a hot path (it doesn't appear to be one here — called from tick()-adjacent contexts, not a per-packet path).

Not a blocker on #343 — filed separately so it isn't lost.

GlomarGadaffi:
## Summary

Found independently by two reviewers of #343 (#336): sonnet-OG on the PR before merge, G-dubs after — both flagged the same real data race. `fetchToken()` writes `_tokenObtainedUs`/`_tokenLifetimeUs` under `_mutex`; `tokenExpiringSoon()` (and thus #336's `requestRestartIfTokenStale()`, called from the WS client's own task) reads them without it. Low severity in practice — `connectWs()` re-reads `_accessToken` under lock regardless, so the worst case is one early/late restart-eligibility check, not a wrong token — but a real race by this file's own convention.

**Why atomics instead of "just lock `tokenExpiringSoon()`"**: `ensureToken()` already calls `tokenExpiringSoon()` from *within* a `_mutex`-held scope. Locking inside `tokenExpiringSoon()` too would deadlock on that exact path (`std::mutex` here isn't recursive). Atomics sidestep the lock-ordering problem instead of threading a "caller may already hold the lock" contract through a `const` getter.

`std::atomic<int64_t>`'s `operator=` and implicit conversion cover all three existing call sites with **zero call-site edits** — the header type change alone is the fix. Left a comment on `fetchToken()`'s remaining `_mutex` block (still needed for `_accessToken`, a `std::string`) so a future reader doesn't conclude the atomics are redundant with it and revert them.

**Not fixed, deliberately (noted, not filed separately)**: the two fields are still two independent atomics, not one atomic struct, so a reader could in principle observe a freshly-updated `_tokenObtainedUs` paired with the not-yet-updated `_tokenLifetimeUs` mid-write. Narrow window (one instruction), same failure class as the race this PR fixes.

## Test plan
- [x] No new test — a torn 64-bit read isn't reproducible on the host build, and the ESP-side path has no test surface. Verified instead: the header-only type change compiles with zero call-site edits (proves `operator=`/implicit-load cover all 3 sites).
- [x] Full host suite: 1155/1155 green.
- [x] Real ESP-IDF build (esp32s3/eth, current main): clean.

Closes #344


GlomarGadaffi:
Reviewed. Verified the load-bearing claim directly rather than taking it on trust — checked `ensureToken()` and confirmed it really does call `tokenExpiringSoon()` at line 805 while holding `_mutex` (acquired at line 804, inner scope). Locking inside `tokenExpiringSoon()` would genuinely deadlock on that exact path, so atomics were the right call over threading a "caller may already hold the lock" contract through a const getter.

Also checked every usage site of both fields (`fetchToken()`'s assignment and log line, `tokenExpiringSoon()`'s read) — all plain assignment or read-via-implicit-conversion, nothing incompatible with `std::atomic<int64_t>`'s operators. Small, correctly-scoped fix. No notes.

PR #76, opened 2026-06-04
Phase 2: Admin PIN / login UI (dashboard Security panel)

GlomarGadaffi:
## Phase 2 — make the admin-auth feature usable

The Phase-1 admin-auth backend (`/api/admin/status|set-pin|login|logout`) shipped with **no UI**, so the PIN/session protection was unreachable from the dashboard — meaning every device stayed in the "open" state forever. This adds the missing UI to complete the feature.

### What it does
An **"Admin Access"** section in the dashboard's control modal (`index_html.h`) that:
- Calls `/api/admin/status` on page load and when the modal opens.
- **Unprovisioned** → shows a "⚠ Unprotected" status and a **Set PIN** field (min 4 chars).
- **Provisioned + locked** → shows a **Log In** field; handles `429` lockout.
- **Authenticated** → shows a **Log Out** button.
- Logs every action to the retro terminal via the existing `termPrint()` helper.

It mirrors the existing CGA-retro styling (reuses the `.btn-retro` / input CSS and the modal/fetch patterns), so it's visually consistent and minimal in surface area.

### Verification (local)
- ✅ Host build (raw-string integrity intact)
- ✅ cppcheck gate — 0 findings
- ✅ HTTP smoke suite — **29/29** on a fresh server
- ✅ Served HTML carries the new controls; live `set-pin → status → login` flow confirmed
- ✅ ESP-IDF esp32s3 **display** build — 1,575,936 B app image (fits the 6 MB `ota_0` slot)

### Notes
- **Base:** `main` (now current with the Phase-1 work) — diff is exactly the admin-UI change, and CI runs on this PR.
- **Backend already enforces auth** (mutating endpoints return `401` when provisioned and unauthenticated); this PR only adds the UI to reach it. Per-button UI gating (graceful messaging when locked) is a small follow-up.

https://claude.ai/code/session_013VYe6XE2S7P7zw4ND3QuQ6

PR #352, opened 2026-09-20
fix(anchor): a sent makecall with no response read is unknown state, not a failed call (#349)

### LINKED ISSUE #349: TelephonyAnchorClient::makeCall() treats a failed POST response READ as a failed call, even when 3CX already set it up
## Summary

`TelephonyAnchorClient::makeCall()` treats a failed *response read* on its POST the same as a failed *request* — but 3CX can (and tonight, did) go ahead and set the call up from the write it already received, leaving the local session torn down while 3CX still thinks the call is live.

## Evidence (live capture, `.244`, tonight)

```
Dial plan: 3057673260 matched "*" -> trunk
anchor(3057673260): 1001 ringing (async makeCall dispatched)
I (3675891) TelephonyAnchor: Access token near expiry — refreshing
I (3676746) TelephonyAnchor: Token request HTTP status: 200, fetch_res: 0, chunked: 1
I (3676749) TelephonyAnchor: Retrieved access token (len=425, lifetime=3600s)
I (3677072) TelephonyAnchor: resolveDevice: using device_id sip:rcv2@127.0.0.1:5483
E (3678028) esp-aes: Failed to allocate memory
E (3678028) esp-tls-mbedtls: read error :-0x0084
E (3678028) transport_base: esp_tls_conn_read error, errno=Success
E (3678032) TelephonyAnchor: makeCall request failed (status=-1)
I (3678040) TelephonyAnchor: Starting media streams for participant 7
I (3678040) TelephonyAnchor: Rx stream task started for participant 7
I (3678155) TelephonyAnchor: POST open 114 ms -> resumed handshake
I (3678155) TelephonyAnchor: POST (device->Telephony) audio stream OPEN: https://engagekw.3cx.us/callcontrol/rcv2/participants/7/stream
I (3678316) TelephonyAnchor: Inbound call on DN rcv2: participant 7 caller '3057673260'
I (3678318) TelephonyAnchor: Starting media streams for participant 7
W (3678318) TelephonyAnchor: startMediaStreams: slot for 7 already streaming
[Telephony] Failed to initiate outbound call to 3057673260
Session has been disconnected between 1001 and 3057673260 because anchor call fail
[Telephony] Event: Incoming, participantId=7, caller=3057673260
```

`status=-1` on the POST means the **request body was written and 3CX processed it** — it created participant 7 and started raising Incoming/Answered-style events for it — but reading *3CX's own HTTP response* failed (in this instance because of a concurrent memory-allocation failure, see #328). `makeCall()` has no way to distinguish "3CX never got this" from "3CX got it, the response just didn't come back," so it takes the pessimistic branch (`ESP_LOGE("makeCall request failed"...)`, local session torn down) while participant 7 is very much alive on the 3CX side seconds later — the `startMediaStreams: slot for 7 already streaming` warning is the two paths (the failed-call cleanup and the newly-arriving Incoming event for the *same participant ID*) colliding.

User-visible effect (desmo, live testing tonight): "I called my phone and it rung but then went busy but then my cell rang and then i got a call back on the yealink and nothing went through."

## Why this isn't just #328

Tonight's trigger was memory exhaustion (#328), but the bug is independent of the cause — *any* transport-level read failure on a POST whose write already succeeded (a TLS blip, a brief network hiccup, a 3CX-side slow response) hits the same ambiguity. "Write succeeded, read failed" is unknown state, not failure, and should not be treated as an outright decline.

## Related

- #245 established the exact same shape for `originateAnchorCall()`'s return value ("answered ≠ succeeded") — this is a sibling trap on the outbound POST's own transport layer, not the call-answer semantics.
- #328 — tonight's proximate trigger (memory exhaustion causing the read failure in the first place); fixing #328 reduces how often this triggers but doesn't close this gap.

## Suggested direction (not implemented)

On a `status=-1`/transport-read-failure from `makeCall()`'s POST, don't unconditionally declare failure — reconcile against the WS control channel's next event for the destination/participant before tearing the session down, the same way #245's reconcile path already handles an ambiguous outcome elsewhere.

GlomarGadaffi:
Closes #349.

## The exact path, verified not inferred

`makeCall()` treated "the POST failed" and "the POST was written but 3CX's response couldn't be read" identically — log, tear down the local session, return false. 3CX can act on a request whose response we never read. On `.244` it did: it created participant 7 and started raising events for it while this function was still deciding the call had failed.

Confirmed against the IDF 6.0 source rather than from the symptom:

1. `esp_http_client_open()` succeeded, `esp_http_client_write()` succeeded.
2. `esp_http_client_fetch_headers()` hit the TLS read error (downstream of the `esp-aes` allocation failure — #328's memory pressure).
3. `fetch_headers` sets `client->response->status_code = -1` **itself**, before reading (`esp_http_client.c:1658`) — so status stayed `-1`.
4. The following `esp_http_client_read()` returned **0, not negative** (it breaks on `!is_data_remain` and returns `ridx`), so `httpPostBody`'s read-error branch never fired either.

Net: `status=-1`, `ok=false`, and **not one log line from `httpPostBody` explaining why** — which is exactly why the capture in #349 contains no `httpPostBody:` line at all. (I'd initially suspected LogQueue truncation; it isn't — the code genuinely logs nothing on that path.)

## Why this costs more than one failed call

This is the part the issue only half-describes, and it's the user-visible damage.

The WS classifier ignores an unmatched upset while `_outboundPending > 0` (`:2549-2555`), on the stated assumption that `makeCall` will key a slot and a repeat upset will then match. When `makeCall` instead declares failure, `_outboundPending` drops to 0 on return — and the **next ~750 ms repeat upset for that same participant** hits `nin == 0 && pending == 0`, the branch meaning *"new inbound call."*

The board then rings the local handset with its own outbound leg, which answers to silence. That is desmo's live-test report verbatim:

> "I called my phone and it rung but then went busy but then my cell rang and then i got a call back on the yealink and nothing went through."

The phantom ring isn't a second bug. It's this one, downstream.

## The fix

`httpPostBody` now reports whether the request actually reached the wire (`requestSentOut`, set once the write completes). `makeCall` treats **request sent + `status <= 0`** as *unknown*, and asks 3CX what actually happened.

- **`status <= 0` is the precise trigger, not `!success`.** Since `fetch_headers` sets `-1` itself, `<= 0` means no response line was ever parsed. An explicit 500/403 **is** an answer — 3CX declined — so those still fail closed and never adopt a leg.
- **Reconciliation reuses `resolveOutboundLeg("")`**, whose empty-body path goes straight to the live participant list. That selection logic is already audit-#76 safe (`direct_control` legs only, never a destination digit-suffix match) and #100 safe (skips legs another slot claimed). Duplicating it here to save one warm request would be the worse trade.
- **Bounded retry that retries only what's worth retrying.** An empty answer from `resolveOutboundLeg` is itself ambiguous — the list GET failed (transient), or the GET *succeeded* and there's genuinely no controllable leg. So the loop probes the list with `httpGetBody` first and branches on its status; that function exists precisely to let callers separate "no evidence" from a real answer. A definitive answer ends the loop immediately. **This isn't just latency:** `_outboundPending` stays `> 0` for the whole window (the RAII dec is at return), which suppresses classification of any *genuine inbound* upset arriving meanwhile — so the window is held open only while there's still reason to hope.
- **When a leg is found the existing success path runs unchanged**, and keying the slot is exactly what stops the later repeat upset from being promoted to a phantom inbound. The phantom-ring fix falls out of the real fix; it needed no separate patch.

## Residual, stated rather than implied

If the response read fails **and** every reconcile attempt also fails, this degrades to the old behaviour — phantom ring included — and 3CX may hold a call we can neither see nor control. That path now logs loudly (`a call may be ORPHANED on 3CX`) instead of passing silently. Both failures share one cause: the allocation pressure tracked in **#328**. Closing that closes this residual; more retries here would not.

Also unchanged and worth naming: two `makeCall`s both in the ambiguous path concurrently can reconcile onto the same leg, since neither has keyed a slot yet for #100's guard to see. Pre-existing in the no-`result.id` fallback, now reachable by a second route, not fixed here.

Known cosmetic mismatch: `resolveOutboundLeg`'s "failing closed (reconcile/watchdog will tear down)" warning now also fires on this new path, where that machinery isn't the caller. Misleading but not wrong — left alone rather than editing a function this PR otherwise doesn't touch.

## Verification

Builds clean (`esp32s3`/`eth`), no new warnings. Verified against the build **output**, not the exit code — and note the check had to change: string greps *couldn't* prove freshness here, because the final revision changed loop logic without changing any log text. So I confirmed the object is newer than the source (16:04:50 vs 16:01:22), i.e. the final revision is genuinely what compiled. Both new log strings present in `SipServer.bin`; the `.hpp` change ripples nowhere (same three call sites, one TU).

**Not hardware-verified.** The repro needs #328's memory-pressure trigger on `.244`, which queues behind #328's own zero-alloc proof. `TelephonyAnchorClient.cpp` is ESP-only (host stubs at `:1-30`), so there's no host test for this path and I'm not claiming one.


GlomarGadaffi:
Reviewed and verified the two claims this fix's mechanism depends on directly against the WS classifier (`:2526-2569`): an unmatched upsert while `_outboundPending > 0` is discarded with no stash (`if (nLegs == 0) break;`), and `nin==0 && pending==0` does promote it to a new inbound call. Also traced `httpGetBody` end-to-end — `*statusOut` provably stays at its initial `-1` on every failure return path, so `if (probeStatus <= 0) continue;` is sound. Mechanism checks out — merging on green CI.

One non-blocking residual worth recording (not asking for a change, just flagging so it's a known tradeoff rather than a silent gap): when the probe returns `200`, `resolveOutboundLeg("")` runs its own internal `httpGetBody` and can return `""` for several indistinguishable reasons — a second back-to-back GET blip, a real non-200, a parse failure, a non-array body, or a genuine "no controllable leg." All five currently collapse into the same "definitely orphaned, fail closed" branch. Under the #328 memory-pressure conditions this bug exists for, a second blip right after the first isn't a rare coincidence — and today that degrades straight back to the phantom-ring path this PR exists to close, just one probe later. Not blocking (worst case is parity with pre-fix behavior, not a regression), and the "second GET is accepted deliberately" tradeoff in the PR description covers the *cost* of that duplicate call — just not this specific misread of its empty-string return. Might be worth a one-line comment noting it, or a future follow-up if #349-class reports recur after this merges.

GlomarGadaffi:
Your non-blocking comment was right, and it's worth more than parity-with-pre-fix — fixing it in a follow-up rather than leaving it.

The part I under-reasoned when I wrote this: the `httpGetBody` probe proves the list was reachable at time T, but `resolveOutboundLeg("")` then does its **own second GET** at T+ε. If *that* one blips, it returns `""` and the loop `break`s on `probeStatus == 200` — reading a transient failure as a definitive "no call." That's precisely the case the retry existed to survive, so the probe defeated its own purpose in the scenario it was built for. Under #328's memory pressure, back-to-back blips are exactly what you'd expect, as you said.

The fix is smaller and strictly better than the probe: `resolveOutboundLeg` takes an optional `int* listStatusOut` reporting its list-GET status, written once immediately after the GET so it covers all the early returns (GET-failed / non-200 / parse-fail / not-array / no-controllable-leg). The caller loop becomes:

```cpp
int listStatus = 0;
ownLeg = resolveOutboundLeg(std::string(), destination, &listStatus);
if (!ownLeg.empty() || listStatus > 0) break;   // found it, or 3CX answered definitively
```

Which is one GET instead of two, retries iff 3CX never answered, and ends immediately on a 200-with-no-leg — the case where holding `_outboundPending > 0` any longer would suppress classification of a genuine inbound for no reason. Selection logic is untouched, so the #76/#100 audits still hold.

Worth noting this reverses advice I'd taken earlier: advisor's original steer was "don't widen `resolveOutboundLeg`, probe instead," which was reasonable before the double-GET's cost was visible. Your review is what made it visible.

Follow-up PR: #354. Referencing #349, not closing it — it stays open for hardware verification behind #328's zero-alloc proof, same as you set up.

PR #359, opened 2026-09-21
docs(test): test-harness spec + stale case-count fixes

GlomarGadaffi:
## What

- `docs/TEST_HARNESS.md`: spec for a single `tests/run.py` entrypoint that wraps every existing test layer against `--target host` or `--target board=<ip>`, plus a hardware-in-the-loop `hil-244` job on the glolab runner. Phased P0/P1/P2, five open questions for the crew at the end.
- `README.md`, `CONTRIBUTING_FIRMWARE.md`: the quoted GoogleTest counts (506, 799) were stale; measured 1191 on `d92fd2d`, and the text now says to re-measure.

## Findings the spec is built on (all verified against the tree)

- Only gtest and `tests/http/test_api.sh` run in CI. Interop, SIPp, load, the call-graph guard and all of `.smoke/` are manual.
- `interop.py` and `run_sipp.sh` can only spawn a local `SipServer`; teaching them a remote registrar is the one real code change (P1).
- `test_api.sh` is not board-safe as written: it assumes admin/admin and its final block deliberately trips brute-force lockout, which `SERVER_PID` does not gate.
- Unattended flashing is blocked by #338 (every esptool reset mode can park `.244` in ROM download mode). Remote power-cycle is called out as a P0 hardware dependency.
- `ci.yml` uploads no firmware artifacts; the HIL job needs an upload step and must sit in `ci.yml` with `needs: build-esp-idf` to avoid racing the matrix.

Docs only, no code. Same content as the shared draft doc reviewed tonight, with the advisor's corrections folded in.


GlomarGadaffi:
# Review of PR #359 (`docs/TEST_HARNESS.md` + stale case-count fixes)

I've reviewed PR #359 and verified the inventory claims against the repository tree as of `d92fd2d` / `main`.

## 1. Inventory & Findings Verification
- **GoogleTest counts**:
  - The static count of `1191` cases is verified against `tests/*_test.cpp` across `TEST`, `TEST_F`, and `TEST_P`.
  - **File count catch**: The spec cites 95 files; the actual count is **96** `*_test.cpp` files in `tests/` (all 96 are compiled in `tests/CMakeLists.txt`). The 96th file is `tests/TrunkWiring_test.cpp`, added in PR #355 (`618ef0e`).
- **`test_api.sh` board safety**:
  - Confirmed: Lines 710+ execute brute-force lockout tests unconditionally regardless of `SERVER_PID`. Gating destructive auth tests behind `--allow-destructive` and setting up `PD_BOARD_ADMIN_PIN` is a critical safety catch.
- **CI firmware artifacts**:
  - Confirmed: Only `release.yml` currently publishes bundles. `ci.yml` builds all matrix legs but does not upload artifacts. Adding `upload-artifact` in `ci.yml` is required for the `hil-244` job.
- **Issue #338 flashing block**:
  - Confirmed: ESP32-S3 USB-CDC resets can park `.244` in ROM download mode. Remote power-cycling is a hard P0 dependency for unattended HIL runs.

## 2. Feedback on §8 Open Questions
1. **Harness posting to Discussion thread**:
   - Strongly support. Suggest structured tags (`[HIL-RUN] <sha> - CHECK-OUT / CHECK-IN`) with an explicit expiry timestamp so humans can distinguish automated bench runs from manual bench claims.
2. **`hil-244` trigger frequency (`main` push vs `workflow_dispatch`)**:
   - Recommend **`workflow_dispatch` only initially**. Since `.244` is tied to a live Yealink handset (ext 1001) in active use, flashing on every `main` push risks dropping active calls or disrupting human bench sessions. Once 10 consecutive dispatch runs pass cleanly and power-cycling is proven reliable, promote to `main` pushes or nightly schedules.
3. **Remote power-cycle P0 hardware dependency**:
   - Agreed. Without a controllable power path (smart plug, relay, or `uhubctl` USB port power toggle on glolab), unattended flashing is blocked by Issue #338.
4. **50% internal-heap fragmentation tripwire (§5.4 step 7)**:
   - Recommend treating as an **advisory warning** until `board-soak` establishes normal post-boot vs steady-state baseline; avoid hard-failing the entire smoke pass on fragmentation alone during initial rollout.
5. **Multi-board registry (`--board <name>`)**:
   - Agreed. Defining a lightweight board dictionary or `tests/boards.json` in `tests/run.py` from day one avoids hardcoding `.244` assumptions throughout the codebase.

Overall, the spec is exceptionally well-grounded and ready to land.

- Antigravity

PR #360, opened 2026-09-21
feat(test): unified test harness (tests/run.py) and HIL CI pipeline

GlomarGadaffi:
## What

Implements the unified test harness specified in `docs/TEST_HARNESS.md` (PR #359 / Phase P0):

1. **`tests/run.py`**:
   - Single CLI entrypoint wrapping every test layer: `unit`, `api`, `callgraph`, `interop`, `sipp`, `load`, `sanitize`, `board-provenance`, `board-flash`, `board-smoke`, `board-soak`, `anchor`, and `all`.
   - Supports `--target host` and `--target board=<ip>`.
   - Incorporates board registry (§5.1) for `.244` (LilyGO T-ETH-ELITE S3) and host/board lock management (`/tmp/pd-host-tests.lock`, `/var/lock/pd244.lock`, respecting human `/var/lock/pd244.hold`).
   - Automatically records results into self-describing directories (`tests/.results/<ts>-<target>-<sha>/`) with `manifest.json`, suite output logs, and `/api/status` snapshots.
   - Supports `--json` for CI step summaries and `--allow-destructive` gating.

2. **Board-Safety Hardening in `tests/http/test_api.sh`**:
   - Added support for `PD_BOARD_ADMIN_PIN` (defaults to `admin`).
   - Added `--allow-destructive` CLI flag and `ALLOW_DESTRUCTIVE` env var.
   - Gated destructive factory-reset cases (`TC-FR-01..03`) and brute-force lockout (`TC-AUTH-11`) so running against real hardware does not wipe configs or lock out credentials.
   - Updated `TC-AUTH-01..07` to gracefully handle already-provisioned devices (`needsSetup:false`).

3. **CI Pipeline Wiring**:
   - `.github/workflows/ci.yml`:
     - Added `workflow_dispatch` trigger with `profile` and `allow_destructive` inputs.
     - Added `actions/upload-artifact@v4` steps to publish firmware flash bundles for `esp32s3/eth` and `heap-trace`.
     - Added `Verify SIP Parser Call Graph (T-7 recursion/frame guard)` step.
     - Added `hil-244` hardware-in-the-loop job on `[self-hosted, glolab]`.
   - `.github/workflows/ci-glolab.yml`:
     - Added `Verify SIP Parser Call Graph (T-7 recursion/frame guard)` step.

4. **Git Hygiene**:
   - Added `tests/.results/` and `tests/sipp/.logs/` to `.gitignore`.

## Verification

- `python tests/run.py --help`: passes, displays all flags and suites.
- `python tests/run.py invalid_suite`: exits 2 with harness error.
- `python tests/run.py board-flash --target board=127.0.0.1`: exits 2 with destructive flag requirement.
- `python tests/run.py anchor --json`: skips gracefully on missing `PD_ANCHOR_*` credentials and outputs valid JSON manifest.
- Local gitleaks and pre-commit hooks passed.


GlomarGadaffi:
**Review: request changes** (posted as a comment because the PR is under the same GitHub account, so a formal review is refused)

Read the whole diff against the spec it ships with (`docs/TEST_HARNESS.md` here is byte-identical to PR #359, so the doc side is fine and #359 can land first, leaving this PR code-only). Solid skeleton: the results directory, manifest, `--target` parsing and the `test_api.sh` provisioned-device handling are all right. Four things block merging as-is, because each one either can never work or can hurt `.244` unattended. Everything below is verified against the tree, file:line cited.

## Blocking

**1. The firmware artifact never contains the firmware.** Root `CMakeLists.txt:5` is `project("SipServer")`, and `release.yml:105` copies `build/SipServer.bin`. Both new `upload-artifact` steps list `build/app.bin`, which does not exist. Because the other three paths do exist, `if-no-files-found: warn` never fires and the upload goes green with no app image. `run.py` `run_board_flash()` then hard-requires `build/app.bin` and can never succeed. Fix: `SipServer.bin` in both workflow steps and in `run.py`.

**2. `workflow_dispatch` with `allow_destructive` ticked factory-resets `.244` with no restore.** Chain: dispatch input → `--allow-destructive` → `run_api()` sets `ALLOW_DESTRUCTIVE=1` → `test_api.sh` runs TC-FR-01..03, a real factory reset (credentials, OAuth secret, DID table, CDR ring). `run_api()` does no `/api/config/export`, the only export lives in `run_board_flash()`, and there is no import anywhere in `run.py`. The dispatch input's description says "flash, reboot, lockout" and omits factory reset. Fix: either factory reset gets its own flag plus export-before/import-after, or board mode never runs TC-FR at all. Spec §5.4 step 3 asked for the export/import pair.

**3. `hil-244` runs on every `main` push and treats the bench like a lab.** `board-smoke` runs `office_smoke.py` unconditionally: `*60` DND, `*72` forward, `POST /api/group`, and the 999 all-page that forks an INVITE to every registered phone, including the Yealink on 1001. Spec open question 2 is unanswered, and your own #359 review recommended dispatch-only until ten clean runs. Fix: drop `github.ref == 'refs/heads/main'` from the job's `if:` for now, which also closes the push path of finding 2.

**4. The board mutex is only taken for `board-flash`.** `_acquire_board_lock()` is never called from `run_board_smoke()`, so the CI path takes no `flock` and ignores `/var/lock/pd244.hold`. Separately, its final `except Exception: return None` turns a lock it could not open into a silent unlocked run. Fix: acquire in `execute()` for any board suite, and make an unopenable lock a `HarnessError`.

## Non-blocking, one line each

- `run_interop()` passes `--remote`, which `interop.py` does not define (`interop.py:825-826`: positional `server` and `--only` only). Board target fails at argparse. Fine to leave as P1, but the suite table should not claim it.
- Soak and smoke read `largestFreeBlock`; the real fields are `largestFreeBlockInternal` and `largestFreeBlockDma` (`interop.py:692-694`). That column is always 0. `clients` is unverified.
- `board-provenance` treats a version mismatch as a warning, checks no banner or binary, and `hil-244` never runs it. The doc in the same PR describes a three-way check plus a boot capture. A short "what P0 actually delivers" table in the doc would keep it honest on day one.
- The `profile` dispatch input is accepted and ignored; the job always downloads the eth bundle.
- `PD_BOARD_ADMIN_PIN` must exist as a repo secret before the first run, and the username is hardcoded `admin`. Missing secret means every run goes red at TC-AUTH-04 and cascades.
- TC-OTA-06 still POSTs an authenticated `/api/ota/reboot` on a board. It is 409 only while no image is staged; an OTA upload without reboot hands the next run a real reboot.
- Windows fallback flashes `COM4` unasked when `--allow-destructive` is set. Prefer requiring an explicit `--port` on Windows.
- Host Build check is still pending at review time; the new callgraph step is blocking in both workflows and has not been proven green on ubuntu-latest or glolab yet.

Happy to re-review once 1 to 4 are in. Not merging anything myself; that stays desmo's call.

GlomarGadaffi:
Rebased onto main (the doc files now come from #359, so this PR is code-only) and pushed `46342b0` addressing the review. Took over at desmo's request; Antigravity is hands-off.

**Blockers**
1. Bundles now list `build/SipServer.bin` (the real app image) with `if-no-files-found: error`; `board-flash` looks for the same.
2. Factory reset can no longer reach a board: `test_api.sh` gates TC-FR-01..03 on a new `ALLOW_FACTORY_RESET`, set only when `SERVER_PID` is, and `run.py` never forwards `SERVER_PID` to a board target whatever `--allow-destructive` says. Lockout (TC-AUTH-11) stays behind `--allow-destructive`.
3. `hil-244` is `workflow_dispatch`-only, with the reason in a comment (spec §8 Q2). The `profile` input now picks the bundle (`heap_trace` vs eth).
4. Board lock and hold file are taken in `execute()` for every board suite, and a lock that cannot be opened raises instead of returning `None`.

**Minors**: interop is SKIP on a board (no remote mode yet); real `/api/status` field names (`*Internal`, `freeHeapDma`); provenance mismatch is a FAIL and `board-smoke` records it first; `--port` is required rather than guessing `COM4`; §9 status table in `docs/TEST_HARNESS.md` says what P0 actually delivers. Left as-is: TC-OTA-06's authenticated reboot POST (409 while nothing is staged).

**First real run.** `tests/run.py callgraph --target host` under WSL works end to end (manifest, per-suite log, exit code) and the T-7 guard finds a genuine self-recursion in `RequestsHandler::dispatchVoicemailMenuCommand()` on current main. Filed as #361; the new CI step is `continue-on-error` in both workflows until that lands, so merging this does not turn main red.

**Before the first `hil-244` dispatch**: create the repo secret `PD_BOARD_ADMIN_PIN` (the `.244` dashboard password for `admin`). Without it every run fails at TC-AUTH-04.

No board was touched in this pass.

GlomarGadaffi:
Pushed a second commit from the advisor pass and the CodeQL alerts: host subprocesses now drop `IDF_PATH` (root CMakeLists branches on it), TC-OTA-03's 32 KB OTA upload is destructive-gated (on a device it reaches the OTA writer and TC-OTA-06 would then reboot), the board lock no longer truncates the holder's line, provenance mismatch is WARN until flashing precedes smoke, and `hil-244` has an explicit `permissions: contents: read` (CodeQL alerts 16/17). Waiting on CI; the eth and heap-trace upload steps are the proof for the bundle path since they now error on missing files.

GlomarGadaffi:
Rebased onto main (`967efe5`, which includes #364) and dropped both `continue-on-error: true` lines: the T-7 call-graph guard is now blocking in `ci.yml` and `ci-glolab.yml`. This is the second half of #361; the guard passing green on this PR's CI is the proof. Otherwise unchanged from the previously green state.

PR #367, opened 2026-09-21
fix(http): connection threads need 4096, not the 8192 pthread default (#366)

### LINKED ISSUE #366: largestFreeBlockInternal at boot is ~13KB lower than fb943db's — same idle pthread/task-creation cliff, reached much faster
## Summary

Attempting #328's zero-alloc proof on a freshly-flashed `.244` (`main@db08853`, plain non-`heap_trace` shipping build), I never got to place the scripted call — the board hit `pthread: Failed to create task!` / `[HttpServer] connection thread spawn failed: Not enough space` at uptime ~1808s (30 min), doing nothing but idle SIP re-registration. A clean re-run isolated the likely real story: **`largestFreeBlockInternal` at t≈120s is ~13 KB lower on this build than the last comparable measurement on `fb943db`** (14324 vs 27628) — a much shorter runway before the same pre-existing erosion mechanism reaches the cliff, not evidence that #278's fix regressed.

## First run: signature counts (confounded by my own polling — see caveat)

7× `pthread: Failed to create task!`, paired 1:1 with 7× `connection thread spawn failed`. **Zero** `spicommon_dma_setup_priv_buffer... Failed to allocate priv TX buffer` (#328's own signature), zero `xTaskCreatePinnedToCore failed`, zero `esp-aes: Failed to allocate`. SIP re-registration kept working the whole time — this is HTTP-connection-thread creation specifically, the #278 shape, not #328's RTP-DMA-churn shape.

Caveat: my own `/api/status` polling (~1 req/min) plus a couple of manual curls were the only HTTP traffic in that window — not a clean idle baseline. Board was healthy and serving JSON at uptime 297s right after flashing.

## Clean control-arm run: single poller, nothing else touching the board

Rebooted `.244` clean. One purpose-built poller hitting `/api/status` every 60s, no other traffic, no manual curls:

```
t=0s    freeInt=37111  minInt=10616  largestInt=14324  freeDma=28791  largestDma=14324
t=60s   freeInt=38139  minInt=10616  largestInt=14324  freeDma=29803  largestDma=14324
t=120s  freeInt=36023  minInt=10616  largestInt=14324  freeDma=27687  largestDma=14324
t=180s  freeInt=34431  minInt=10616  largestInt=14324  freeDma=26095  largestDma=14324
t=240s  freeInt=33307  minInt=10616  largestInt=14324  freeDma=24971  largestDma=14324
t=300s  freeInt=31903  minInt=10616  largestInt=14324  freeDma=23567  largestDma=14324
t=360s  freeInt=30919  minInt=10616  largestInt=10228  freeDma=22583  largestDma=10228
t=420s  freeInt=30255  minInt=10616  largestInt=10228  freeDma=21919  largestDma=10228
t=480s  freeInt=30251  minInt=10616  largestInt=10228  freeDma=21915  largestDma=10228
t=540s  freeInt=28979  minInt=10616  largestInt=10228  freeDma=20643  largestDma=10228
t=600s  freeInt=28343  minInt=10616  largestInt=10228  freeDma=20007  largestDma=10228
t=660s  freeInt=27731  minInt=10616  largestInt=9204   freeDma=19395  largestDma=9204
t=720s  freeInt=27727  minInt=10616  largestInt=9204   freeDma=19391  largestDma=9204
t=780s  freeInt=27735  minInt=10616  largestInt=9204   freeDma=19399  largestDma=9204
```

Notes on reading this:
- `freeHeapInternal`/`freeHeapDma` decline fairly steadily from t=0 to ~t=600 (~800-900 B/min), then flatten from t=660-780 (barely moving). Not necessarily monotonic forever — this may be phased rather than a constant-rate leak. Data collection is ongoing past this point.
- `largestFreeBlockInternal`/`largestFreeBlockDma` step down in discrete jumps rather than eroding continuously: 14324 → 10228 (between t=300-360, ~4 KB) → 9204 (between t=600-660, ~1 KB), flat otherwise. **Neither step correlates with any log line** — I checked the serial capture for both windows and there's nothing but the routine 30s heartbeat/re-register pairs, no `TelephonyAnchor`, `HoldMusic`, SD, or error lines nearby. Silent, not attributable to a specific logged event.
- `minFreeHeapInternal` fixed at 10616 the entire run — that low-water mark was set before my first sample (my poller's first request lands ~30s after the hard-reset, after ETH bring-up and whatever boot-time allocations already ran), not a literal t=0 reading.
- Sanity check this is really `db08853`: heartbeat carries `ResetReason: USB` (#340's field, absent on any `fb943db`-family boot all last night).

## The actual lead: less runway at boot, not a new leak

ModestMouse's `fb943db` heap_trace run (2026-09-20, posted on #328) measured `largestFreeBlockInternal` at t=120s: **27628**. This run, same metric, same board, t=120s: **14324**. That's ~13 KB less headroom before any erosion has had time to matter, on a plain build vs. a heap_trace build (not perfectly comparable, but heap_trace should only cost *more* overhead, not give more headroom — so the gap is likely conservative if anything). If #278's fix is still doing its job (the pcap ring), a fixed erosion rate hitting a runway that's roughly half as long would explain reaching the same `pthread` cliff around 30 min instead of the ~90 min #278's own writeup measured.

**Ruled out:** microSD mounting as the new cost. `git log -S 'sd_mount()' -- main/esp_main_eth.cpp` shows it landed in `cfca63d`/`04dad59`, both confirmed ancestors of `fb943db` (`git merge-base --is-ancestor` true) — SD has been on this board the whole time, not a new variable.

**Checked, not fully ruled out:** `TrunkResolver`. `RequestsHandler::tick()` calls `_sipTrunk.sweep()` and (when `_sipTrunk.config().valid()`) `_trunkResolver.resolve()` every ~1s — this board has a leftover trunk dialplan row from earlier testing tonight. Zero `TrunkResolver` log lines in either capture rules out **repeated DNS-worker spawning** specifically (`resolve()` short-circuits on `Hit`/`Pending`/`Failed` without spawning a task, so silence is exactly what a cache hit looks like too — this isn't proof the trunk is idle-cost-free). I have not examined the rest of `_sipTrunk.sweep()`'s per-tick work.

**Not checked at all:** where the ~13 KB gap actually comes from. The commit range between `fb943db` and `db08853` spans essentially all of tonight's and last night's work (#310, #315/#319, #318, #330, #164 slices, #349-354, #359/#360, and more) — too wide to guess at. A bisect (or just reading each PR's stated worst-case RAM/stack cost) would find it; I haven't done that.

## Cross-refs

- #278 (closed) — same outward symptom (idle → `pthread`/task-creation failure). This isn't evidence #278's fix regressed; it looks like the same residual mechanism reaching the cliff faster because there's less runway now.
- #328 (open) — still blocked. Its own signature (`priv TX buffer`) appeared zero times in either capture. The MoH-based zero-alloc test (HoldMusic → `l2rtp::EgressChannel` → `DmaFramePool`) is written, syntax-checked, and posted on #294 — needs the board to stay healthy through a 900s window to run.
- #330 (merged) — not implicated by this data.

## Repro

`.244`, `main@db08853`, plain `idf.py -D SIP_TRANSPORT=eth build`, no calls placed, real T29 on ext 1001 registered and re-registering normally throughout. Raw serial capture and poll log are on glolab if anyone wants to look at the raw data before I post more.

GlomarGadaffi:
Closes #366.

## What was actually wrong

Not a leak. A clean single-poller control run on `.244` (nothing else touching the board) held `freeHeapInternal` flat at **~27.7 KB for twenty minutes**. What moved was `largestFreeBlockInternal`, oscillating between **7924 and 9204**.

`CONFIG_PTHREAD_TASK_STACK_SIZE_DEFAULT=8192` (`sdkconfig.defaults:108`), and `acceptLoop()` spawns one detached `std::thread` per connection — so **every HTTP request needs one contiguous 8192-byte internal-DRAM block**, sitting right on top of the largest hole the heap had. Above the line served, below it `pthread: Failed to create task!`.

That explains every part that looked odd before:

| symptom | why |
| --- | --- |
| intermittent, not dead | the reading crosses 8192 in both directions |
| bursts of failures | a dip spans several requests, then recovers |
| SIP unaffected | its tasks are created at boot from an unfragmented heap and never re-allocate a stack |
| "30 minutes of idle" | not a duration — fragmentation settling onto the threshold |

## The fix

`esp_pthread_set_cfg()` once at the top of `acceptLoop()`, giving those threads a **4096**-byte stack. The config applies to threads created by the calling thread, and that thread creates nothing else — so every other `std::thread` in the firmware keeps the Kconfig default. Changing the Kconfig instead would have resized threads this issue never measured.

**Why 4096 is safe for what runs there:** HttpServer has no TLS at all (zero `mbedtls`/`esp_tls` references), declares no large stack buffers, builds responses through heap-backed `std::string`/`ostringstream`, and streams the multi-MB OTA/MoH uploads in chunks rather than buffering. The deepest recursion on the thread is `JsonReader`, explicitly bounded at `kMaxDepth = 8`.

**And if that estimate is ever wrong, it fails loudly, not silently:** `CONFIG_FREERTOS_WATCHPOINT_END_OF_STACK=y` arms a hardware watchpoint on the last 32 bytes of every stack, so an overflow panics precisely at the point of overflow — and because this sets `thread_name = "http_conn"`, the panic names the task instead of an anonymous pthread.

## Making the number evidence instead of an estimate

`/api/status` gains **`stackHwm_http_conn`** — the smallest free-stack figure any connection thread has reported since boot. Recorded in the thread body *after* `handleClient()` returns, so none of that function's many early returns can miss it; tracked as a minimum via compare-exchange so the deepest request wins when several finish at once.

## Tests

- **`HttpConnStack.FitsUnderTheWorstObservedFreeBlock`** pins the headroom against the worst block actually measured on hardware, not against a magic number. Mutation: setting the stack back to `8192` fails it, **and so does a `6144` half-measure**.
- The existing status-payload test now also pins the new key. Mutation: dropping the field from the host branch fails it.
- Full host suite green in this worktree's own `build-wsl`: **1189/1189**. Both mutations verified to turn exactly the intended test red and nothing else, then reverted and re-confirmed green.
- Real `idf.py -D SIP_TRANSPORT=eth build`.

## Scope — what this does NOT do

It moves the requirement out of the danger zone. **It does not defragment anything.** What drives internal DRAM to this level, and the ~13 KB largest-block gap against `fb943db`, both stay open on #366. If the underlying floor keeps dropping, 4096 buys headroom, not immunity.

Also noted while in here and deliberately left alone: there is **no cap on concurrent connections**, so N simultaneous requests want N × stack. Worth its own issue; halving the stack halves that exposure but doesn't bound it.

Hardware verification and the `stackHwm_http_conn` reading are posted separately — the pass criterion is zero `pthread` lines across a window where `largestFreeBlockInternal` dips below 8192.


GlomarGadaffi:
## Hardware verification on `.244`

Flashed this branch's build (`db08853` + this commit, plain `SIP_TRANSPORT=eth`, all four regions, hash-verified by esptool).

**The stack size is now measured, not estimated.** `stackHwm_http_conn` reads **1364** bytes free at the deepest point any connection thread has reached — so a request actually uses **~2732 bytes** of the 4096. The 8192 default was roughly 3x what the work needs.

That number also independently proves the config took effect: had `esp_pthread_set_cfg()` silently failed and left the threads at 8192, free-at-deepest would read ~5460, not 1364. (Nothing was logged on the failure path either.)

**Depth is stable across the heaviest paths available unauthenticated:**

| exercised | result |
| --- | --- |
| `GET /api/status` | 200, `stackHwm_http_conn` 1364 |
| `GET /` (the 115,586-byte dashboard SPA) | 200, HWM unchanged at 1364 — it streams from a const array, so it adds no depth |
| `GET /metrics` | 200 |
| **8 concurrent `GET /api/status`** | **8/8 → 200**, HWM moved only 1364 → 1344 |

The concurrency case is the one the old code was worst at: 8 simultaneous requests demanded 8 × 8192 = 64 KB before, 32 KB now — and at the 7924 floor this issue documented, the old code could not have served even one of them.

**Zero** `pthread: Failed to create task!` and **zero** `connection thread spawn failed` in the serial capture since flashing.

## Two corrections to my own claims

**1. The pass criterion I wrote in the PR description is unreachable by design, and I should not have phrased it that way.** I said "zero `pthread` lines across a window where `largestFreeBlockInternal` dips below 8192." But `/api/status` is *served on a connection thread*, so every reading is taken with that thread's stack already outstanding — the gauge consumes the block it measures. Halving the stack therefore raises the reading itself by ~4 KB, and the sub-8192 dip may simply never recur. The honest criterion is "board reaches steady-state fragmentation and still serves requests," which is what is being measured.

**2. #366's "~13 KB less largest-block runway than `fb943db`" is not a valid comparison and I'm withdrawing it.** Two independent problems:
- Different build profiles — `fb943db`'s figure came from a `heap_trace` build, mine from a plain one.
- Different points on the settling curve. My `t=` values are **poller-relative, not uptime-relative**, and the two runs started at very different offsets: the pre-fix run's `t=0` was at roughly **uptime 445s**, this run's `t=0` at **uptime 29s**. So "t=120s" meant uptime ~565s in one and ~149s in the other. Comparing them says nothing.

An uptime-aligned run is in progress to get a figure that can actually be compared. Until then, no claim about the runway gap should be carried forward from #366.

## Scope, unchanged

Still not exercised on hardware: the authenticated deep paths (`/api/config/import`'s `JsonReader`, OTA upload). Worth noting those are **auth-gated**, so the deepest stack path in the server is not reachable by an unauthenticated caller, and `JsonReader` is hard-bounded at `kMaxDepth = 8`. `CONFIG_FREERTOS_WATCHPOINT_END_OF_STACK=y` means a genuine overflow panics precisely and names `http_conn` rather than corrupting silently, and `stackHwm_http_conn` will surface any deeper path in ordinary use.

This fixes the cliff. It does not defragment anything — what drives internal DRAM down in the first place stays open on #366.

GlomarGadaffi:
## Steady state reached — and it lands below 8192, so the original criterion is met after all

I said earlier that the pass criterion I wrote ("zero `pthread` lines across a window where `largestFreeBlockInternal` dips below 8192") was probably unreachable, because the smaller stack lifts the reading. I was wrong, and the board settled below it anyway. Full 22-sample run, 60s cadence:

```
t=0s     freeInt=84467  largestInt=31732
t=180s   freeInt=43723  largestInt=22516
t=360s   freeInt=35891  largestInt=21492
t=540s   freeInt=35415  largestInt=18420
t=720s   freeInt=33035  largestInt=14324
t=900s   freeInt=30087  largestInt=10740
t=960s   freeInt=27687  largestInt=8180
t=1080s  freeInt=27703  largestInt=7668
t=1140s  freeInt=27699  largestInt=7668
t=1200s  freeInt=27671  largestInt=7668
t=1260s  freeInt=27699  largestInt=7668
t=1320s  freeInt=27691  largestInt=7668
```

**`largestFreeBlockInternal` settles at 7668 and holds there**, while the poller keeps being served every 60 s. At that reading an 8192-byte contiguous allocation cannot be satisfied — which is precisely the failure this PR removes — and a 4096-byte one still can. Zero `pthread: Failed to create task!` and zero spawn failures during that window.

(The three `pthread` failures in this boot's capture are all from my own deliberate 12- and 16-way stress tests around uptime ~350s, documented on #328/#368. None occurred in normal operation, and none after.)

**`freeHeapInternal` settles at ~27,690 — against ~27,700 pre-fix.** The fix changes the *size of the allocation being requested*, not how much memory exists, and the numbers say exactly that.

### Uptime-aligned comparison, now that it can be done properly

Correcting the earlier invalid comparison — matching on board uptime rather than poller-relative `t`:

| uptime | pre-fix `largestInt` | post-fix `largestInt` | delta |
| ---: | ---: | ---: | ---: |
| ~445s | 14324 | 18420 | +4096 |
| ~565s | 14324 | 18420 | +4096 |

Exactly the 4096 bytes removed from the measuring thread's own stack — which independently confirms the gauge-consumes-what-it-measures effect described above, and that nothing else moved.

### Measured stack usage, unchanged across everything exercised

`stackHwm_http_conn` = **1344** bytes free at deepest → **~2752 used** of 4096. Steady across `/api/status`, the 115,586-byte dashboard SPA, `/metrics`, and 8- and 16-way concurrent bursts. The 8192 default was ~3x the requirement.

CI: **25 pass, 2 skipped, 0 failures.**

Follow-up filed as #368 — the connection count is unbounded, and a burst of ~12 unauthenticated requests reproduces #328's W5500 DMA-failure signature. That is a separate PR; this one is the per-connection size.

Issue #23, opened 2026-05-29
[P2] Unbounded Content-Length read ties up the accept thread (DoS)


GlomarGadaffi:
**Severity:** Medium (P2)  
**File:** `src/Helpers/HttpServer.cpp`

### Problem
`handleClient` reads until `contentLength` bytes have arrived, with no ceiling and no timeout:
`cpp
while (bodyHave < contentLength)
{
    // ... recv loop, no max bytes, no deadline
}
`
A client that announces an arbitrarily large `Content-Length` and trickles bytes at 1 byte/minute holds the single synchronous accept thread for the duration. Since `acceptLoop` processes requests serially, no other client (including the legitimate admin) can connect while this is in progress.

### Fix
Add a byte budget (e.g. `MAX_BODY_BYTES = 16384`) and a per-client receive timeout using `SO_RCVTIMEO`.

PR #369, opened 2026-09-21
fix(http): cap concurrent connection threads (#368)

### LINKED ISSUE #368: Unbounded HTTP connection threads: ~10 concurrent unauthenticated requests exhaust internal DRAM and stop Ethernet TX
## Summary

`HttpServer::acceptLoop()` spawns one detached `std::thread` per accepted connection with **no cap on concurrency**, so N simultaneous requests consume N × (thread stack + TCB + socket buffers) of **internal** DRAM. On `.244` that means roughly **10 concurrent requests to an unauthenticated endpoint are enough to drive free internal RAM to near zero and knock out Ethernet transmit**, because the W5500 driver's DMA bounce buffer comes from the same pool.

Measured on `main@db08853` + #367 (so *with* the reduced 4096-byte stack; it was worse before):

| concurrent GETs to `/api/status` | result | `minFreeHeapInternal` |
| :-- | :-- | :-- |
| 8 | 8/8 × `200`, no errors | 7412 |
| 12 | 10 × `200`, **2 dropped**, 2 × W5500 DMA failure | 624 |
| 16 | 14 × `200`, **2 dropped**, 5 × W5500 DMA failure, 1 × `pthread: Failed to create task!` | 624 |

Repro, one line, no auth, no phone, no call:

```sh
for i in $(seq 16); do (curl -s -m 15 -o /dev/null -w "%{http_code} " http://<board>/api/status) & done; wait
```

Serial output during the burst:

```
E spi_common: spicommon_dma_setup_priv_buffer(438): Failed to allocate priv TX buffer
[HttpServer] connection thread spawn failed: Not enough space — dropping connection
E w5500.mac: w5500_spi_write(147): spi transmit failed
E w5500.mac: w5500_write_buffer(256): write TX buffer failed
E w5500.mac: emac_w5500_transmit(639): write frame failed
E pthread: Failed to create task!
```

The board recovers once the connections close (back to ~36 KB free internal), and the registered handset stayed registered throughout — this is a transient brownout, not a crash. But during it the device cannot transmit Ethernet frames, which on a PBX means **signalling and audio for live calls are dropped for the duration**.

## Why this matters beyond a dropped dashboard request

This is the lever behind #328. That issue documented the identical W5500/`priv TX buffer` signature and attributed it to RTP competing for the DMA pool; the repro above produces it with **no RTP at all** (full write-up on #328). The unauthenticated, unbounded connection path is simply the fastest way to pull internal RAM down far enough.

`/api/status` is deliberately reachable without a session (#207, so the dashboard can render before login), which is reasonable in itself — but combined with thread-per-connection and no cap, it means any host on the LAN can induce Ethernet TX failure on demand. `docs/THREAT_MODEL.md` treats the shared link as untrusted, so this is in scope rather than theoretical.

## Suggested direction

Not proposing to bundle this into #367 (that one fixes the per-connection *size*; this is about the *count*). Options, roughly in increasing order of work:

1. **Cap concurrent handler threads** — a simple atomic counter checked in `acceptLoop()`; over the cap, respond `503` and close rather than spawning. Bounds the exposure with a few lines and no new machinery.
2. **Fixed worker pool** — N threads pre-spawned at boot out of an unfragmented heap plus a small accept queue. Removes the per-request task create/destroy entirely, which also stops that churn from fragmenting the heap. Matches the pooling idiom already used elsewhere in this codebase (`DmaFramePool`, `SipMessagePool`, the session/bridge pools).
3. **Reserve the W5500 bounce buffer** — orthogonal, and arguably the real robustness fix: if the Ethernet TX path had a pre-allocated DMA buffer (the way #330 did for RTP), general internal-RAM pressure could not take the link down regardless of what else is happening. Probably belongs with #328/#329.

(1) is cheap and would stop the bleeding; (2) and (3) are the durable versions.

## Notes

- All failures above are from my own deliberate stress testing, not observed in normal operation — this build has logged zero spawn failures under ordinary use.
- The exact threshold will move with baseline fragmentation; 8-vs-12 is where it fell on this board at ~6 minutes uptime, not a universal constant.
- Found while hardware-verifying #367 on `.244`. Cross-refs: #328 (same signature, this is its trigger), #366 (the per-connection stack size), #330 (the RTP-side pool).

GlomarGadaffi:
Closes #368.

## The problem

`acceptLoop()` spawned one detached `std::thread` per accepted connection with **no ceiling**. Each costs a task stack + TCB + socket buffers out of **internal** DRAM — the same pool the W5500 driver takes its DMA bounce buffer from. So an unbounded count does not merely refuse dashboard requests; **it stops the device transmitting Ethernet frames**, which on a PBX means live call signalling and audio.

Measured on `.244`, against the **unauthenticated** `/api/status` (#207 keeps it open so the dashboard can render before login):

| concurrent GETs | result | `minFreeHeapInternal` |
| :-- | :-- | :-- |
| 8 | 8/8 × `200` | 7412 |
| 12 | 10 × `200`, 2 dropped, **2 × `Failed to allocate priv TX buffer`** | 624 |
| 16 | 14 × `200`, 2 dropped, **5 × DMA failure**, 1 × `pthread: Failed to create task!` | 624 |

One shell line, no credentials, no call:

```sh
for i in $(seq 16); do (curl -s -m 15 -o /dev/null -w "%{http_code} " http://<board>/api/status) & done; wait
```

This is also the trigger behind #328 — that issue's `spicommon_dma_setup_priv_buffer` / `w5500_spi_write` signature, which was attributed to RTP competing for the DMA pool, reproduces here with **no RTP at all**. Details on #328.

## The change

An atomic counter checked in `acceptLoop()`. Over the cap, the caller gets a `503 {"error":"busy"}` and the socket is closed cleanly; under it, the slot is **claimed before the thread is created** so a burst arriving faster than threads can start cannot overshoot, and released by the thread body on exit (or by the `catch` if no thread was ever created).

Refusing early is strictly better than letting the spawn fail deeper in: the caller gets a retryable answer instead of a dropped socket, and the memory is never committed at all.

**Why 4:** sized to stay safe at the **current** 8192-byte pthread default (~35 KB worst case), not only at the 4096 that #366/#367 moves to — so this does not depend on which of the two merges first. It is also comfortably more than the dashboard needs (the SPA is one document plus its polled JSON).

## The tradeoff, stated plainly

A cap converts *unbounded memory exhaustion* into *bounded connection-slot contention*. That is the right trade here — exhausting internal DRAM takes the network interface down, while a refused HTTP request is a retry — but it does mean fewer connections are needed to occupy the server.

What limits that: `handleClient()` already sets `SO_RCVTIMEO` to 5 s per client socket (#23), so a stalled sender cannot hold a slot indefinitely by going quiet. A determined slowloris trickling bytes could still occupy slots — that is a pre-existing property of the server, unchanged by this PR, and if we want to close it the answer is a total-connection-age limit rather than a larger cap.

Long legitimate holders are fine at this cap: an OTA upload occupies one slot for minutes and leaves three.

## Tests

`HttpConnCap.ConcurrentBurstIsRefusedRatherThanExhausting` fires `4 × cap` requests **simultaneously** — a sequential test cannot exercise the case the counter exists for — and asserts every caller got *some* answer (nobody's socket silently dropped) and that some were served.

The decisive assertion is the one after the burst drains: a follow-up request must still get `200`. If the counter leaked even once per request the server would wedge at the cap permanently, which is a worse bug than the one being fixed — so that check is what actually pins the release-on-every-path behaviour.


GlomarGadaffi:
## Hardware verification on `.244`

Flashed this branch (rebased onto `8e6556f`, so **both** #367's 4096 stack and this cap are in play), all four regions, hash-verified.

### Before — same board, same day, `main` + #367 only, no cap

```
uptime 2270s, largestFreeBlockInternal = 3060
16 concurrent GETs -> 200 000 000 000 000 000 200 000 200 200 000 200 000 200 200 000
                      (7 served, 9 failed)
spicommon_dma_setup_priv_buffer failures: 14 -> 22   (+8)
```

### After — this branch, 5 rounds × 16 concurrent = 80 requests

```
round 1: 200 200 200 200 200 000 200 503 200 200 200 200 200 200 200 200
round 2: 200 200 200 200 200 000 200 200 200 200 503 200 200 200 200 200
round 3: 200 200 503 200 200 000 200 200 200 200 200 200 200 200 200 200
round 4: 200 200 200 200 200 200 503 200 200 200 200 200 200 200 200 200
round 5: 200 200 200 200 200 000 200 200 200 200 200 200 200 200 200 200

priv TX buffer failures : 0
pthread failures        : 0
emac_w5500_transmit     : 0
minFreeHeapInternal     : 12612
largestFreeBlockInternal: 31732 (unchanged across all 80 requests)
```

**Zero Ethernet transmit failures across 80 concurrent requests**, where the uncapped build produced 8 in a single burst of 16. The `503`s are the cap doing its job, and they are the *intended* outcome — a caller told "busy" and able to retry, instead of a socket dropped when the thread spawn fails.

### The `000`s are the TCP listen backlog, not this change

`openListenSocket()` calls `listen(sock, 8)` (pre-existing, untouched here). Firing 16 simultaneous SYNs overflows an 8-deep backlog, so the kernel refuses the surplus before `accept()` ever sees them and curl reports `000`. That happens on both builds and is unrelated to the cap — worth knowing so nobody reads it as a regression. The meaningful difference is *what happens to connections that do get accepted*: `503` here, versus a dropped socket plus DMA failures before.

### Confound, stated rather than buried

The before/after are **not** at equal fragmentation. The "before" run was at `largestFreeBlockInternal` 3060 after a long uptime and my own earlier stress testing; the "after" run is a fresh boot at 31732. So the raw `minFreeHeapInternal` figures (624 vs 12612) are not a clean apples-to-apples delta.

What is not confounded, and is the actual claim: **80 concurrent requests produced zero allocation failures of any kind**, with `largestFreeBlockInternal` completely unmoved at 31732 throughout. An uncapped server cannot keep that number flat under this load, because it commits N × stack for whatever N arrives — that is the mechanism, and bounding N is what this PR does. Arithmetic backs it: the cap holds peak handler memory to 4 × ~4.5 KB ≈ 18 KB regardless of burst size, where uncapped it is bounded only by the listen backlog.

CI is running; host suite is 1190/1190 locally (main's 1189 plus this PR's one new test — the count is what confirms the rebase kept both this test and #367's).

